PR #204, opened 2014-11-16
Adds head, tail, and isEmpty methods to several objects

yloiseau:
This PR adds `head`, `tail`, and `isEmpty` methods on several collection-like objects:
* adds augmentation on `CharSequence`,
* changes the behaviour of augmentation on `List`,
* adds native (java) methods to `Tuple`,
* adds native (java) methods to ranges,
* adds special methods to arrays.

It also introduce a java interface, `gololang.HeadTail<T>`, and makes `Tuple` and ranges to implement it.

These methods behaviour is:
* `isEmpty` tells if the container is empty,
* `head` returns the first element, or `null` if the collection is empty, to allow chaining methods with “elvis” operator `?:`,
* `tail` returns a new object of the same type containing the rest of the elements (or an empty object if empty).

The purpose is to provide to all collection-like (immutable) structures an homogeneous interface suitable for recursive algorithms.

As an example, a generic function computing the size of such an object can be:
```golo
function size = |struct| -> match {
  when struct: isEmpty() then 0
  otherwise 1 + size(struct: tail())
}
```
and the sum of all the elements:
```golo
function sum = |struct| -> match {
  when struct: isEmpty() then 0
  otherwise struct: head() + sum(struct: tail())
}
```
we can manipulate strings with the same aproach:
```golo
function upper = |str| -> match {
  when str: isEmpty() then ""
  otherwise toUpperCase(str: head()) + upper(str: tail()
}
```


yloiseau:
Can we merge this one ?

jponge:
Let's have it for 2.1.0 :-)

src/main/java/gololang/HeadTail.java
```diff
@@ -0,0 +1,62 @@
...
+ * }
+ * </pre>
+ * <p>
+ * Note that the {@code size} method is not provided since this interface
+ * can be implemented by infinite generators.
+ * <p>
+ * A {@code List} augmentation is provided, as well as corresponding special
+ * methods on arrays ({@code Object[]}/{@code array[]})
+ */
+public interface HeadTail<T> {
```
jponge: You should have javadoc tags for the return values and the parametric type `T` :smiley:

jponge:
Good to merge now I guess.

PR #230, opened 2015-01-03
Fix doc

k33g:
- Add clarification about `andThen`
- Fix sample

### DIFF
--- a/doc/closures.asciidoc
+++ b/doc/closures.asciidoc
@@ -267,7 +267,7 @@ return functions, as in:
 
 [source,golo]
 ----
-let f = |x| -> |y| -> |z| -> -> x + y + z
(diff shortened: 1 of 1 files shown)


yloiseau:
composition using `andThen` isnot limited to closures. You can use any function reference.

```golo
function foo = |x| -> x + 1
function bar = |x| -> 2 * x

let newFunction = ^foo: andThen(^bar)
```

yloiseau:
Maybe “you can compose function and closure” woud be better.

BTW, to be pedantic, in the sample following the sentence, it's just anonymous functions, not closures, since no value is closed in the definition :wink:

k33g:
argh you're right! :)

k33g:
perhaps, this is the sample that isn't complete enough

PR #236, opened 2015-01-17
Create promise with thread

k33g:
# define promise
    let do_it =  -> promise(): initializeWithThread(|resolve, reject| {
       # the code runs inside a thread
      resolve(42)
    })

    # run promise
    do_it(): onSet(|result| { # if success
      println(result)
    }): onFail(|err| { # if failed
      println(err: getMessage())
    })


jponge:
BTW it would be a bit better to make PR from your repo, not the main one :smile:

jponge:
Shouldn't it be called `initializeWithinThread`?

k33g:
oups sorry!

PR #239, opened 2015-01-18
Initialize promise within a Worker

k33g:


src/main/golo/async.golo
```diff
@@ -85,6 +85,17 @@ augment gololang.concurrent.async.Promise {
     }): start()
     return this: future()
   }
+
+  ----
+  You can define a promise which runs "stuff" inside a Worker
```
jponge: Sentences shall end with `.`

src/main/golo/async.golo
```diff
@@ -85,6 +85,17 @@ augment gololang.concurrent.async.Promise {
...
+  You have to pass a `WorkerEnvironment` to `initializeWithinWorker` method
+  ----
+  function initializeWithinWorker = |this, env, closure| {
```
jponge: You should provide a list description of the arguments.

src/main/golo/async.golo
```diff
@@ -85,6 +85,17 @@ augment gololang.concurrent.async.Promise {
...
+    env: spawn(|message| {
+      this: initialize(closure)
+    }): send("")
```
jponge: Why send an empty string?

k33g: What can I send?
null generate an error
and I can't find a use case where the message is useful

Issue #241, opened 2015-01-20
Get rid of compilation warnings


jponge:
A few warnings appear while compiling Golo. We should be able to easily get rid of them, even at the cost of mindful `@SuppressWarning` annotations :smile:

jponge:
Just realized it was already fixed.

PR #242, opened 2015-01-20
Suppress warnings in tests.

yloiseau:
See #241

The warnings in tests compilation were due to:
- tests on deprecated Predefined golo functions;
- raw casts in range tests.

In both cases, it is safe to suppress them.

Also make some code cleaning.


jponge:
Seems like it broke on JDK 8 :-)

jponge:
:+1:

PR #244, opened 2015-01-22
add 'invalid' token

guillaumesoldera:
add an 'invalid' token to avoid TokenMgrError when " or - characters are typed in netbeans. Check not null parameter to avoid NullPointerException when user add an import statement


jponge:
Sounds good

jponge:
Thanks :smiley:

PR #246, opened 2015-02-28
Makes Tuple object comparable.

yloiseau:
`Tuple` now implements `Comparable<Tuple>`. The comparison is done
pairwise, using lexicographical order, such that `[1, 2] < [1, 3]`
and `["b", 5] < ["d", 1]`.

### DIFF
--- a/src/main/java/gololang/Tuple.java
+++ b/src/main/java/gololang/Tuple.java
@@ -142,7 +142,9 @@ public final class Tuple implements Iterable<Object>, HeadTail<Object>, Comparab
    */
   @Override
   public int compareTo(Tuple other) {
-    if (this.equals(other)) {return 0;}
+    if (this.equals(other)) {
+      return 0;
+    }
     if (this.size() != other.size()) {
       throw new IllegalArgumentException(String.format(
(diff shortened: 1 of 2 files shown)


src/main/java/gololang/Tuple.java
```diff
@@ -127,6 +127,36 @@ public boolean equals(Object o) {
...
+   *
+   * @param other the tuple to be compared.
+   * @return a negative integer, zero, or a positive integer as this tuple is less than, equal to, or greater than the specified tuple.
+   * @throws NullPointerException if the specified tuple is null
+   * @throws ClassCastException  if the type of the elements in the specified tuple prevent them from being compared to this tuple elements.
+   * @throws IllegalArgumentException if the specified tuple has a different size than this tuple.
+   */
+  @Override
+  public int compareTo(Tuple other) {
+    if (this.equals(other)) {return 0;}
```
jponge: As a matter of style, could you apply a consistent formatting?

```java
if (plop) {
  return 0;
}
// (...)
```

jponge:
Good to go, just see my comment on formatting.

Issue #219, opened 2014-12-05
Calling an anonymous function


yloiseau:
It's not possible to call directly an anonymous function
```golo
let r = (|a| -> a): invokeWithArguments(1)
println(r)
```
display 1 which is ok, but
```golo
let r = (|a| -> a)(1)
println(r)
```
displays `MethodHandle(Object)Object` which is weired.

Even worse, 
```golo
println((|a| -> a): invokeWithArguments(1))
```
displays 1, but
```golo
println((|a| -> a)(1))
```
raises a syntax error...

PR #247, opened 2015-03-01
Fix Bug #219.

yloiseau:
As described in #219, it was impossible to directly call a anonymous function, as in
`(|x| -> x + 1)(1)`

This is fixed by allowing `AnonymousFunctionInvocation` in an
`AssociativeExpression` in the parser.

This change generates `BinaryOperation` with empty operator that must
thus be checked.


jponge:
All good :+1:

Issue #19, opened 2013-04-21
Add a join method on the List Std Augmentation


danielpetisme:
Cause is always useful!!

Issue #18, opened 2013-04-18
golodoc


jponge:
The language should provide a dedicated syntax for API documentations.

I propose a notation such as:

```
----
Integer and string concatenation. This is *awesome*.

- `a` must be some number
- `b` must be a string

It returns the concatenation of `a` and `b`, as a string.
----
function concat = |a, b| -> a + b
```

The nice thing is that documentation could be written in Markdown, and that the special `----` delimiters could only apply to `function`, `module`, `augment` or any upcoming top-level declaration.

`----` is inspired by blocks in AsciiDoc.

The parser could enforce that, meaning that `#` is for comments while `----` is for documentation.

See #1

Issue #17, opened 2013-04-18
Maven incremental compilation


jponge:
The Golo Maven plugin recompiles everytime it is being invoked. We could improve it by making it incremental. Data comparisons should be enough because Golo is a dynamic language, and there is no cross-module check at compile time.

yloiseau:
Couldn't we close this issue, since we moved to gradle?

jponge:
We still have the Maven plugin, and it doesn’t have incremental compilation :-)

Issue #16, opened 2013-04-18
JVM bytecode debug symbols


jponge:
The Golo bytecode currently lacks debug informations. This makes it hard to spot the source of, say, an exception as there is no correspondence between bytecode and line numbers in the original Golo source code.

PR #250, opened 2015-03-03
Named arguments

danielpetisme:
First of all a bit of vocabulary, a *parameter* (or _formal parameter_) representents the identifier used during the function definition
```golo
function min = |a, b| -> ...
```
`a` and `b` are *parameters*
An *argument* (or _actual parameter_) represents the value used during the function invocation
```golo
min(1,2)
```
`1` and `2` are *arguments*

So far, Golo runtime relied on the arguments order to find the target function/method to invoke. With named parameters, the developers do not have to care about the argument order, plus it permits to increase the verbosity of the function invocations.
```golo
function join = |delimiter, values...| ->

#change argument order + varargs support
join(values = array["Foo", "Bar"], delimiter=",")
```

[Show me more!](https://github.com/golo-lang/golo-lang/blob/bdbcad0caa81a44558377f9cfb2cdc8edf37443f/doc/functions.asciidoc#named-parameters)


yloiseau:
:+1:

doc/functions.asciidoc
```diff
@@ -118,6 +118,38 @@ then:
...
+----
+
+NOTE: If you name one argument named then you must name all the remaining argument. Otherwise, a compilation error will be raised.
+
+Once you are using named parameters in your function call, the order doesn't matter.
+
+[source,golo]
+----
+function dummy = |who, when, what| {
+  return who + " " when + " " + what
```
yloiseau: a `+` is missing between the `" "` and the `when`

Artpej: `when` is a reserved kework and can't be a variable name

danielpetisme:
Doc fixed according to @yloiseau @Artpej comments.
Thank you guys

jponge:
This sounds awesome @danielpetisme :smiley: 

I will dive into the pull-request soon and make a proper code review.

danielpetisme:
The aim is to have a first code review. Once this PR merged, I will have a look on optional parameters

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -48,7 +48,7 @@
   static {
     String bootstrapOwner = "fr/insalyon/citi/golo/runtime/FunctionCallSupport";
     String bootstrapMethod = "bootstrap";
-    String description = "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;I)Ljava/lang/invoke/CallSite;";
+    String description = "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite;";
     FUNCTION_INVOCATION_HANDLE = new Handle(H_INVOKESTATIC, bootstrapOwner, bootstrapMethod, description);
```
Artpej: I think that BSM method doesn't allow parameters of type  `Object` ( [reference](http://asm.ow2.org/asm40/javadoc/user/org/objectweb/asm/MethodVisitor.html#visitInvokeDynamicInsn(java.lang.String,%20java.lang.String,%20org.objectweb.asm.Handle,%20java.lang.Object...)) )

jponge: You can have trailing objects in a bootstrap method, and they can be of `Object` type. The only thing is that they must be compatible with constant pool entries, so primitives, classes, strings and field/method references.

jponge:
@danielpetisme Haven't checked the code yet, but does the resolution checks parameters order for *all* call sites, or just the call sites where parameter names have been used?

danielpetisme:
During the compilation
=================

*Function definition* 
a Java [@GoloFunction](https://github.com/danielpetisme/golo-lang/blob/wip/named-arguments/src/main/java/fr/insalyon/citi/golo/runtime/GoloFunction.java) annotation is generated to decore the Java method (for ALL the golo functions, could be the entry point for more metadata) 
````
function plop = |foo, bar| 
# generates
@GoloFunction(parameters = {"foo", "bar"})
public static Object plop(Object, Object)
````

*Function invocation*
if the invocation is flagged as _using name parameters_, then the Arguments visitor build the list of the argument names used in the current invocation. These arguments are passd to the indy BSM.

Runtime
======
The Function Call Support looks for a method which match the actual arguments constraints (number + order + type). The magic part is that Golo function only use `Object` has type, so that's why we also find the method even if we permute arguments, the validation is made on the number of arguments only.

Then, if the call site store argument names (ie. not empty), we first check if the method is annotated `@GoloFunction` (ie. we know the formal parameters names) and then generate a `permuteArguments` handler
*Voilà*

I hope it answers your question @jponge

**Comment by jponge** [comment]
@danielpetisme That answers 200% my question :smile: :beers:

File: doc/functions.asciidoc [hunk 3]
```diff
@@ -118,6 +118,34 @@ then:
...
+  tags = array["feature", "syntax"],
+  body = "it rocks!"
+  title = "Named parameters are alive",
+  promoted = true
+)
+----
+
+NOTE: If you name one argument then you must name all the remaining argument. Otherwise, a compilation error will be raised.
+
+Once you are using named parameters in your function call, the order doesn't matter anymore.
```
jponge: Should be a note, too.

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
jponge: So the `I` argument gets into the objects array, right?

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -450,14 +462,29 @@ public void visitThrowStatement(ThrowStatement throwStatement) {
...
+    return argumentNames;
   }
 
   @Override
   public void visitFunctionInvocation(FunctionInvocation functionInvocation) {
+    String name = functionInvocation.getName().replaceAll("\\.", "#");
+    String typeDef = goloFunctionSignature(functionInvocation.getArity());
+    Handle handle = FUNCTION_INVOCATION_HANDLE;
+    List<Object> bootstrapArgs = new ArrayList<>();
+    bootstrapArgs.add(functionInvocation.isConstant() ? 1 : 0);
```
jponge: Ok, got my answer for the point above :smile:

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -468,20 +495,12 @@ public void visitFunctionInvocation(FunctionInvocation functionInvocation) {
     if (functionInvocation.isAnonymous() || functionInvocation.isOnReference() || functionInvocation.isOnModuleState()) {
       methodVisitor.visitTypeInsn(CHECKCAST, "java/lang/invoke/MethodHandle");
       MethodType type = genericMethodType(functionInvocation.getArity() + 1).changeParameterType(0, MethodHandle.class);
-      visitInvocationArguments(functionInvocation);
-      methodVisitor.visitInvokeDynamicInsn(
-          functionInvocation.getName().replaceAll("\\.", "#"),
-          type.toMethodDescriptorString(),
-          CLOSURE_INVOCATION_HANDLE,
-          functionInvocation.isConstant() ? 1 : 0);
-    } else {
```
jponge: Haven't we lost a branch here?

src/main/java/fr/insalyon/citi/golo/compiler/ir/FunctionInvocation.java
```diff
@@ -60,6 +61,14 @@ public void setConstant(boolean constant) {
     this.constant = constant;
   }
 
+  public boolean useNamedArguments() {
```
jponge: `usesNamedArugments()` would be more appropriate.

src/main/java/fr/insalyon/citi/golo/compiler/ir/NamedArgument.java
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.insalyon.citi.golo.compiler.ir;
+
+public class NamedArgument extends ExpressionStatement{
```
jponge: Space before the `{`

src/main/java/fr/insalyon/citi/golo/compiler/ir/NamedArgument.java
```diff
@@ -0,0 +1,49 @@
...
+    return expression;
+  }
+
+  public void setExpression(ExpressionStatement expression) {
+    this.expression = expression;
+  }
+
+  @Override
+  public void accept(GoloIrVisitor visitor) {
+
```
jponge: Not going through the expression?

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
```diff
@@ -69,8 +73,13 @@ public static Object samFilter(Class<?> type, Object value) {
     return value;
   }
 
-  public static CallSite bootstrap(Lookup caller, String name, MethodType type, int constant) throws IllegalAccessException, ClassNotFoundException {
-    FunctionCallSite callSite = new FunctionCallSite(caller, name.replaceAll("#", "\\."), type, constant == 1);
+  public static CallSite bootstrap(Lookup caller, String name, MethodType type, Object... bsmArgs) throws IllegalAccessException, ClassNotFoundException {
+    boolean constant = ((int)bsmArgs[0]) == 1;
+    String[] argumentNames = new String[bsmArgs.length - 1];
+    for (int i = 0; i < bsmArgs.length -1; i++) {
```
jponge: Couldn't we use an intrinsic `Arrays.copy(...)` method?

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
```diff
@@ -69,8 +73,13 @@ public static Object samFilter(Class<?> type, Object value) {
...
+      argumentNames[i] = (String) bsmArgs[i+1];
```
jponge: I like space around operators.

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
```diff
@@ -108,11 +118,15 @@ public static Object fallback(FunctionCallSite callSite, Object[] args) throws T
       Method method = (Method) result;
       checkLocalFunctionCallFromSameModuleAugmentation(method, callerClass.getName());
       types = method.getParameterTypes();
-      if (method.isVarArgs() && isLastArgumentAnArray(types.length, args)) {
+      //TODO improve varargs support on named arguments. Matching the last param type + according argument
```
jponge: Sounds interesting, but could you elaborate on that?

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
```diff
@@ -144,6 +158,27 @@ public static Object fallback(FunctionCallSite callSite, Object[] args) throws T
     }
   }
 
+  private static MethodHandle reorderArguments(Method method, MethodHandle handle, String[] argumentNames) {
```
jponge: Lovely

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
```diff
@@ -244,6 +279,7 @@ private static Object findClassWithStaticMethodOrField(Class<?> callerClass, Str
   }
 
   private static Object findStaticMethodOrField(Class<?> klass, String name, Object[] arguments) {
+
```
jponge: Not sure that blank line has much sense :wink:

src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java
```diff
@@ -1618,4 +1610,32 @@ public void banged() throws Throwable {
     assertThat(result, not((Object)42));
 
   }
+
+  @Test
+  public void test_named_parameters() throws Throwable {
```
jponge: I would also add a test case for the cases where some arguments are named, and some are not.

jponge:
This sounds really really good.

jponge:
In terms of scheduling I would put this in 3.0.0, along with the Java 8 shift + functional interfaces integration.

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
Artpej: @jponge Ok! nice! could be usefull for my futur experiments :wink:

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
danielpetisme: Refactoring
the `visitInvokeDynamicInsn` statement was used twice with different arguments. Now, the arguments are initialized for a standard function invocation and overrided if and only the the conditions are met. 
The consequence is less code and only one `visitInvokeDynamicInsn`.

src/main/java/fr/insalyon/citi/golo/compiler/ir/NamedArgument.java
danielpetisme: Currently no. May it helps to remove the named arguments list building.
https://github.com/golo-lang/golo-lang/pull/250/files#diff-135e9a72efe5f924909a57f244b00de1R472
Plus, change the return type from `List<String>` to `void`

src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java
danielpetisme: https://github.com/golo-lang/golo-lang/pull/250/files#diff-d6d7e547b440d72444c214f0442a6b1aR1642

Maybe throw a `GoloCompilationException` would be cleaner but witch type of problem is this? A `PARSING` problem?

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
danielpetisme: a function call w/ named args + varargs found actually a varargs method but this test also check that the last arg is an array... In a named args context, the varargs arry could be at any place in the function call.
I thought about flagging the varargs parameter in the `@GoloFunction` but don't know if it's worthless. 
So far all the tests run successfully so decided the _good enough impl._ (aka. Quick&Dirty :smile: )

danielpetisme: But I agree is not perfect.

danielpetisme:
To be franck, I didn't looked for the impact on the runtime performance (JMH newbie :smile: ). the 2 main updates on the runtime is the `permuteArguments` wrapping of the handler and the argument reordering. 
*Warning*: so far the named args feature works only for golo functions not for closures (`ClosureCallSupport` hasn' t been modified)

In terms of roadmap, if Golo 3.0.0 will be based on Java8, wouldn' t be easier/cleaner to use the `named argument feature` ?
One merged, I' ll try to have a look on optional parameters (and why not default param.)

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
jponge: :+1:

src/main/java/fr/insalyon/citi/golo/compiler/ir/NamedArgument.java
jponge: Ok, got it.

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
jponge: I have a feeling that vararg arguments in presence of named arguments do not mix well... @yloiseau @Artpej what do you think?

jponge: Or else we should support something like `f(a=1, b=2, 3, 4)` but I'm not sure it's great.

jponge: `f(a=1, b=array[2, 3, 4])`

src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java
jponge: I would create a new enum entry specifically for that problem, and indeed throw a `GoloCompilationException`.

jponge:
I never benchmarked `permuteArguments`, but I would be surprised if it had any impact. By design it's a quite straighforward reordering. But we should definitely add a benchmark case for it.

jponge:
Do you have plans to also modify `ClosureCallSupport`? It would make everything consistent.

jponge:
The problem with Java 8 parameter names through reflection is that `javac` does not have it by default, which is quite stupid.

But by relying on this you could get rid of the `GoloFunction` annotation logic and extend named arguments to all functions and even Java class methods...

jponge:
Perhaps Java 8 parameter names would be best after all, as we could unify named parameters support for Golo and Java virtual and static methods.

src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java
yloiseau: to get inspiration from Python :wink:
given
```python
def f(a, b, c, *args):
    print("got {}, {}, {}, {}".format(a, b, c, args))
```

that is 3 named/positional parameters, and a vararg.

the following calls are valid:
```python
f(1, 2, 3)
f(b=2, a=1, c=3)
f(1, c=3, b=2)
f(1, 2, 3, 4, 5, 6)
```
1. call without varargs with positional arguments: outputs `got 1, 2, 3, ()`
2. call without varargs with named arguments: outputs `got 1, 2, 3, ()`
3. call without varargs, `a` is positional and `b` and `c` named: `got 1, 2, 3, ()`
4. call with positional arguments and varargs: outputs `got 1, 2, 3, (4, 5, 6)`

But the calls
```python
f(a=1, 2, 3)
f(b=2, a=1, c=3, 4, 5, 6)
```
fails. Indeed, once one of the arguments is named, all the *following* arguments *must* be named. So in the first call, even if  `a` is at the “right” place, the other arguments must be named. Compare with 3 above.

It is thus possible to mix named arguments and varargs, but *not* in the same call, since a vararg arguments *must* follows all positional parameters in the function signature, and thus can't follow the named arguments in the call. 
This restriction makes sens to me.

Note also that calling `f(a=1, b=2, c=3, args=(4, 5, 6))` is not valid, since `args` is not actually a keyword parameter for the function, which on the other hand could make sens.

As a side note, the python’s “named varargs” feature could also be realy nice to have:
```python
def g(a, b, *args, **kargs):
    print("got {}, {}, {}, {}".format(a, b, args, kargs))

g(1, 2)
g(c=3, b=2, a=1, d=4)
g(1, 2, c=3)
g(1, 2, 3, 4, 5)
g(1, 2, 3, 4, c=5, d=6)
```

outputs 
```
got 1, 2, (), {}
got 1, 2, (), {'c': 3, 'd': 4}
got 1, 2, (), {'c': 3}
got 1, 2, (3, 4, 5), {}
got 1, 2, (3, 4), {'c': 5, 'd': 6}
```
but here again, all arguments following a named argument must be named.

danielpetisme: The `array[]` approach for named varargs parameter does the job. I mean everybody know that the `...` is replaced by `Object[]` so it makes sense.
I'm not sure `f(a = 1,2,3, b = 4, c = 5)` would be clearer, plus from a parsing standpoint is way harder (at least for me :smile: )

On this particular line of code, the main issue is that currently only method flagged w/ varargs + having an array as last arg are trasnformed to collect all the trailing args into an array (is the job of `asFixedArity`). Without the test I added, the varargs value is an array of arrays (a first one obtained to the `varargs` flag on the method, and a second coming from the `array[]` notation.
 
Now in that case
````
function joiner = |delim, values...| -> ...
joiner(values = array[1,2,3], delim = "-")
````
The joiner method will be flagged as varargs but the collector is in the first position not the last... 
The best option I can see  is to add metadata in the generated annotation. Basically, flag the varags collecor then during the runtime, the above `if` will check that the argument corresponding to the  flagged parameter is an instanceof `Object[]`.

File: src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java (hunk 14 above)
Comment (danielpetisme) [inline src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java]: > There are only two hard things in Computer Science: cache invalidation and naming things.
> 
> -- Phil Karlton

Any idea ? is not a parsing issue, an IR issue ?

**Comment by danielpetisme** [comment]
Regarding `ClosureCallSupport`, ASAP. I just have to focus :smile: I think is just a matter of copy&paste (and I don't like that :cry: )

**Comment by danielpetisme** [comment]
Regarding Java 8 parameter indeed it would be nicer but I'm wondering if it's worthless to cover Java methods... 
The `ArrayList#add` signature is`public boolean add(E e)`.  As you noticed thay had a hard time naming the parameter :smile: 
Without being sarcastic, I agree libs/ramworks developers are much more serious on naming.

File: src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java (hunk 9 above)
Comment (danielpetisme) [inline src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java]: I have to redo my tests but the original array is an instance of `Object[]` and the target one is `String[]`. I'm sure I've tested it and it failed in the casting.
With the loop I'm sure to cast every elements.

I'll check again.

File: src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java (hunk 11 above)
Comment (yloiseau) [inline src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java]: The point of named varargs as in Python is not to get a array of values, but a map [arg name -> value], allowing to function to take named args that are not specified in the function signature (useful e.g. for HOF wrapping other functions)

**Comment by yloiseau** [comment]
As a side note, adding a `bindName` (or alike) method to functions, similar to `bindAt`, but... binding arguments by name instead of position, would be awesome!

**Comment by danielpetisme** [comment]
@yloiseau the first step of a fluent api builder :)
````
function post = |title, body, promoted, tags...| {

#partial approach
postPromoted = ^post: bindName("promoted", true) 
#build approach
post: title("a"): promoted(true): body("b"): tags("syntax", features): invoke()
````

File: src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java (hunk 11 above)
Comment (danielpetisme) [inline src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java]: The map approach was my favorite but harder to implement. You create a type mismatch between the callsite target types and the target function definition. I'm sure it's possible (basically exploding the map into an array with the ordered arguments) but quid about perfs (and I know @jponge is picky on that topic :p).

Comment (jponge) [inline src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java]: Passing arguments in a map blows performance. The approach that you have now (annotation + argument permutation combinator) is both elegant and fast :smiley:

File: src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java (hunk 14 above)
Comment (jponge) [inline src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java]: `Problem.Type.INCOMPLETE_NAMED_ARGUMENTS_USAGE`? (or something like that)

**Comment by jponge** [comment]
I concur with you @danielpetisme. I've had a second thought on the named parameters support for Java/foreign code and it gets a lot messier than sticking with the approach you have now. Because Java methods have types, you would need to resolve on both argument types and names, which complicates everything.

A simpler approach would be to say that we support named arguments for Golo constructs (functions, closures, augmentations), but not foreign / Java code.

**Comment by jponge** [comment]
Thoughts welcome on these discussions, we can take time as it's planned for the release after.

**Comment by yloiseau** [comment]
@danielpetisme :+1: for builder approach too

File: src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java (hunk 11 above)
Comment (yloiseau) [inline src/main/java/fr/insalyon/citi/golo/runtime/FunctionCallSupport.java]: The map in python is only for variable named args, just as vararg in Java creates an array. But indeed, dealing with that can introduce an overhead.

**Comment by jponge** [comment]
@danielpetisme @yloiseau +2

**Comment by danielpetisme** [comment]
_Update_

* Typo fixed 
* Raise a `GoloCompilationException` if an invocation mixes named and unamed args

# Java8 named parameters 
I had a look to the [documentation](https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html). You must provide the `-parameters` option to the compiler to be able to get the parameters names through reflection during the  runtime.

This is an sample of bytecode generated with the `-parameters` options.
```
#javac -p -c -v Test

Classfile /Users/daniel/workspace/with-parameters/Test.class
  Last modified 1 avr. 2015; size 433 bytes
  MD5 checksum c16ec708581bc75ab333784c5352f8e3
  Compiled from "Test.java"
public class Test
  SourceFile: "Test.java"
  minor version: 0
  major version: 52
  flags: ACC_PUBLIC, ACC_SUPER
Constant pool:
   #1 = Methodref          #3.#18         //  java/lang/Object."<init>":()V
   #2 = Class              #19            //  Test
   #3 = Class              #20            //  java/lang/Object
   #4 = Utf8               <init>
   #5 = Utf8               ()V
   #6 = Utf8               Code
   #7 = Utf8               LineNumberTable
   #8 = Utf8               foo
   #9 = Utf8               (Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
  #10 = Utf8               MethodParameters
  #11 = Utf8               firstname
  #12 = Utf8               lastname
  #13 = Utf8               main
  #14 = Utf8               ([Ljava/lang/String;)V
  #15 = Utf8               args
  #16 = Utf8               SourceFile
  #17 = Utf8               Test.java
  #18 = NameAndType        #4:#5          //  "<init>":()V
  #19 = Utf8               Test
  #20 = Utf8               java/lang/Object
{
  public Test();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0       
         1: invokespecial #1                  // Method java/lang/Object."<init>":()V
         4: return        
      LineNumberTable:
        line 3: 0

  public java.lang.Object foo(java.lang.Object, java.lang.Object);
    descriptor: (Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=3, args_size=3
         0: aconst_null   
         1: areturn       
      LineNumberTable:
        line 6: 0
    MethodParameters:
      Name                           Flags
      firstname                      
      lastname                       

  public static void main(java.lang.String...);
    descriptor: ([Ljava/lang/String;)V
    flags: ACC_PUBLIC, ACC_STATIC, ACC_VARARGS
    Code:
      stack=0, locals=1, args_size=1
         0: return        
      LineNumberTable:
        line 9: 0
    MethodParameters:
      Name                           Flags
      args                           
}                    
```

Since Golo's bytecode generation relies on ASM, I'll check if it permits to generate the same kind of information.

This systems will permit to replace the current `@GoloFunction` annotation role. Plus, it will permit to interact with Golo arguments from a Java code in a standard way. Nevertheless, we will face the same kind of limitations. Named arguments will not work with 3rd-party binaries (framework) and I don't think they compiled the JDK8 with the `-parameter` flag (eat your own dog food they said :wink).


# Closure support

Golo's closures are compiled as standard Golo, so they are flagged in the bytecode by the annotation `@GoloFunction`.
During the runtime, they are represented by `MethodHandle`s.  
Since one of the `MethodHandle` objective is to hide the direct access to the target `Method`, I can't retrieve the `@GoloFunction` annotation and as consequence get the parameters' name.

What I don't understand is. Since there is no method resolution for closure, the `MethodHandle` is already bind to a Golo Function/Java Method before the `ClosureCallSupport` is even invoked by `indy`.
So, somewhere before the bytecode generation, the target `Method` is accessible, no ?

Otherwise, posibilities are:
* Document the unfaisability of named args for closures (IBM's logic: if is documented, is not a bug anymore)
* Do some Vodoo Black Magic to workaround the MethodHandle pointer (@jponge is not confident on the faisability)

**Comment by danielpetisme** [comment]
\o/
ASM 5.0 permits to generate the parameter names in the bytecode thanks to [MethodVisitor#visitParameter](http://asm.ow2.org/asm50/javadoc/user/org/objectweb/asm/MethodVisitor.html#visitParameter(java.lang.String, int)).
So I generate standard Java8 bytecode and access to the parameter names through the standard Reflection API.
Regarding the access flag I used the `ACC_FINAL` one (maybe I could add `ACC_SYNTETHIC`?).
Further doc [Obtaining Names of Method Parameters](https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html)

**Comment by Artpej** [comment]
excelent! :+1:

**Comment by danielpetisme** [comment]
Cherry on the cake, since it works in a standard way, if a Java source code has been compiled with the `-parameters` flag, then you can use them in Golo.

```
#Java javac -parameters Printer.java
package io;

public class Printer {
 public static Object print(Object prefix, Object what) {
    System.out.println(prefix +  " " + what);
    return null;
 }
}
```

```
#Golo
module bar
import io.Printer
   
let obj = DynamicObject()
   
function main = |args...| {
  print(what = obj, prefix = ">")
}
```

_I check the JDK bytecode, it hasn't the parameter names embedded :cry: _

**Comment by jponge** [comment]
@danielpetisme superb :smiley: 

BTW you should not use `ACC_SYNTHETIC` since parameter names are real. Or if you use that flag, it shall only be on closure capture arguments.

**Comment by danielpetisme** [comment]
Guys I need your opinions on attaching a state on `FunctionReference`

# Immutable ?
Currently `FunctionReference` is an immutable objet. I've added a `Parameter[] parameters` property but in some case we want to build a FunctionReference on a existing `MethodHandle` and we don't have any reference to the target Method (thus attached parameters). So I can invoke `new FunctionReference(handle, null)` but looks ugly. Plus, the force is strong with `FunctionReference` it has a great potential (ie. a lot of property to add) and I don't want to struggle with `null` flags to know what I can use or not.

# The quick and dirty Map
Adding a `Map<String, Object> properties` property to maintain a constant signature in the `FunctionReference` instantiation. It permits to add feature without updating +50 test files (ping @jponge :wink: ). In the other hand, it force to add a level of indirection `map.get("parameters")` (not elegant).

Thoughts ?

**Comment by jponge** [comment]
@danielpetisme What if you go with the immutable pill, use a `String[]` array for parameter names in order, and just assume that a `null` array means that you don't know?

It is also a much more compact representation than using collection maps.

Bonus point: impact change is simple if you just add a new constructor :smiley: and the one that just takes a method handle delegates to `(handle, null)`.

**Comment by danielpetisme** [comment]
Was my first choice and I think your right but :
* I hate to deal w/ `null`
* Modifications on  `FunctionReference` have a lot of impacts so I tried to anticipate a litle bit (too much? :wink: )

Immutable sounds good for me

**Comment by jponge** [comment]
There are clever ways to deal with `null` ;-)

**Comment by danielpetisme** [comment]
For me the PR is "feature complete". I'm all ears for your feedback.

File: doc/functions.asciidoc [hunk 15]
```diff
@@ -118,6 +118,34 @@ then:
...
+  body = "it rocks!"
+  title = "Named parameters are alive",
+  promoted = true
+)
+----
+
+NOTE: Once you are using named parameters in your function call, the order doesn't matter anymore.
+
+CAUTION: If you name one argument then you must name all the remaining argument. Otherwise, a compilation error will be raised.
+
```
yloiseau: Maybe add a note / explain more precisely how to use named arguments:
- can I `post("the title", "the body", tags=array[...], promoted=false)` since you said “all the *remaining* argumentS”
- in the presence of varargs (clear from the example, but a explaining note may be worth)

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -451,14 +454,26 @@ public void visitThrowStatement(ThrowStatement throwStatement) {
     methodVisitor.visitInsn(ATHROW);
   }
 
-  private void visitInvocationArguments(AbstractInvocation invocation) {
-    for (ExpressionStatement statement : invocation.getArguments()) {
-      statement.accept(this);
+  private List<String> visitInvocationArguments(AbstractInvocation invocation) {
+    List<String> argumentNames = new ArrayList<>();
+    for (ExpressionStatement argument : invocation.getArguments()) {
+      if (invocation instanceof FunctionInvocation && ((FunctionInvocation) invocation).usesNamedArguments()) {
```
yloiseau: Can't we use named arguments on method invocation?

doc/functions.asciidoc
danielpetisme: Such snippet can't work.
Either you name all or none of your arguments. 

For varargs, you expect a note explaining that you have to create an array to box the arguments?

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
danielpetisme: Have a look to the `java-interop`
https://github.com/golo-lang/golo-lang/pull/250/files#diff-73d861b2e4cdd14514d7975925c2b003R65

Basicaly if the bytecode contains the parameters info  it will work.

src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java
```diff
@@ -1689,10 +1718,41 @@ public void banged() throws Throwable {
 
   }
 
-  @Test
   public void lambda8_interop() throws Throwable {
```
yloiseau: no more a test?

doc/functions.asciidoc
yloiseau: Yep, your last commit make the point

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
yloiseau: I mean you put the named argument stuf on `ir.FunctionInvocation` not on `ir.AbstractInvocation`. So, I wonder if given e.g.
```golo
augment java.lang.String {

}
```
```golo
myobject: 
```

yloiseau: Tested... Blam!
```golo
module Test

augment java.lang.String {
  function bar = |this, pref, suf| -> pref + " " + this + " " + suf
}

function main = |args| {
  println("world": bar(suf="!", pref="hello"))
}
```

```bash
$ golo golo --files test.golo
[error] Encountered " "=" "= "" at line 9, column 27.
```

jponge:
I'm adding doing a full review in my backlog :-)

src/test/java/fr/insalyon/citi/golo/compiler/CompileAndRunTest.java
danielpetisme: My bad, fixed

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
danielpetisme: Indeed...
Fixed
https://github.com/danielpetisme/golo-lang/commit/5a82c9156fad4004468389ba302e8249af0f3951#diff-798c54ba96714a3ddf68f1a3db92b6a3R28

As mentionned, I pulled up the named argument stuff into `AbstractInvocation`
https://github.com/danielpetisme/golo-lang/blob/5a82c9156fad4004468389ba302e8249af0f3951/src/main/java/fr/insalyon/citi/golo/compiler/ir/AbstractInvocation.java

yloiseau: :+1:

src/main/java/fr/insalyon/citi/golo/compiler/ir/NamedArgument.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2012-2014 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
```
jponge: 2012-(...)2015 :-)

jponge:
The merge with #265 will require some conflict resolution I'm afraid, but it should not be too complex.

danielpetisme:
Indeed there are some conflicts. I gave an eye on them and it's really not big deal.

These are the differences git can not auto-merge when I try to merge #265 on top of #250.
I tried to analyse the differnces to find a way to permit a nice auto-merge :wink: .

1- In ` src/main/jjtree/Golo.jjt`
Create an `Argument` token for the lexer + Changing the return type of `AnonymousFunctionInvocation()` from `void` to `ASTAnonymousFunctionInvocation ` 
```
<<<<<<< HEAD
void Argument():
{
  Token parameterId = null;
  Token expression;
}
{
  (LOOKAHEAD(2) (parameterId=<IDENTIFIER> "="))? ExpressionStatement()
  {
    if (parameterId != null) {
      jjtThis.setName(parameterId.image);
      jjtThis.setNamed(true);
    }
  }
}

void Arguments() #void: {}
{
  Argument() ("," (BlankLine())? Argument())*
}

void AnonymousFunctionInvocation():
=======
ASTAnonymousFunctionInvocation AnonymousFunctionInvocation():
>>>>>>> Fix LL(k) operators priority
```

I can change the return type on my branch it will works (and permit an auto-merge) but it would be inconsistent with the function usage (in my branch).
For my personal comprehesnsion what this change permits ?

2- In `src/main/java/fr/insalyon/citi/golo/doc/ModuleDocumentation.java`
The word *Argument* is for the expression used in the function invocation and *Parameter* for function declaration. Plus, there is  an Indentation issue

@jponge did you change your IDE spacing settings ? The current [master](https://github.com/golo-lang/golo-lang/blob/master/src%2Fmain%2Fjava%2Ffr%2Finsalyon%2Fciti%2Fgolo%2Fdoc%2FModuleDocumentation.java) has a continuation indent of 2 spaces, [your branch](https://github.com/jponge/golo-lang/blob/fix-260/src%2Fmain%2Fjava%2Ffr%2Finsalyon%2Fciti%2Fgolo%2Fdoc%2FModuleDocumentation.java) has 4 spaces. 

```
<<<<<<< HEAD
          .arguments(node.getParameters())
          .varargs(node.isVarargs());
=======
            .arguments(node.getArguments())
            .varargs(node.isVarargs());
>>>>>>> Fix LL(k) operators priority
```

3- In ` src/main/java/fr/insalyon/citi/golo/compiler/ParseTreeToGoloIrVisitor.java` , s/Parameter/Argument/ + improve the expression parsing
```
<<<<<<< HEAD
    for(String parameter : function.getParameters()) {
      ASTCommutativeExpression commutativeExpression = new ASTCommutativeExpression(0);
      functionInvocation.jjtAddChild(commutativeExpression, functionInvocation.jjtGetNumChildren());
      ASTAssociativeExpression associativeExpression = new ASTAssociativeExpression(0);
      commutativeExpression.jjtAddChild(associativeExpression, 0);
      ASTReference ref = new ASTReference(0);
      ref.setName(parameter);
      associativeExpression.jjtAddChild(ref, 0);
=======
    for (String argument : function.getArguments()) {
      ASTReference ref = new ASTReference(0);
      ref.setName(argument);
      functionInvocation.jjtAddChild(ref, functionInvocation.jjtGetNumChildren());
>>>>>>> Fix LL(k) operators priority
```

Don't know how we could re-arrange the code to auto-merge this one... ::confounded:


The other solution is to merge one PR and fix the issue on the second one. Even if I would prefer an auto-merge, the differences are really no big deal to merge.

jponge:
1. It's because I needed to capture a reference to the element, so no big deal for you (in the end it's just like calling a non-`void` method and ignoring the return value if you don't need it).
2. Perhaps just the nested indent has changed, otherwise I'm on 2-spaces.
3. Basically `ASTCommutativeExpression` and `ASTAssociativeExpression` go away with my changes.

PR #251, opened 2015-03-15
Print JVM version from command line 'version'

rlespinasse:
With the command 'version' and the option '--full',
this print the actual golo full version and the JVM version.

For example :
```
Golo version : 2.1.0-SNAPSHOT (build 15-03-15-12:34)
JVM version: 1.8.0_31
```


jponge:
Sounds good :+1:

PR #252, opened 2015-03-15
Escape module name during project creation to avoid a compilation failure

rlespinasse:
During a project creation with 'new' command line,
if a project name contains a non-word caracter,
they are escape into the dot caracter to avoid a compilation failure.


jponge:
Good catch!

PR #256, opened 2015-03-24
Fix tests and doc of anonymous function call.

yloiseau:
This PR fixes some oversight in PR #247

- adds a few tests on direct anonymous function call when the function
  is returned by a method invocation
- change the doc accordingly, since now the warning in the
  “calling functions that return functions” section is no longer valid


jponge:
:+1:

PR #258, opened 2015-04-06
Sum algebraic data type

yloiseau:
This PR adds sum algebraic data types, also known as tagged union, to Golo
(see <http://en.wikipedia.org/wiki/Tagged_union>).
Product data types are already provided by `struct` and `tuple`.

## Usage example

Some well known usages of sum types are:

### Enumerations

```golo
union Color = {
  RED
  GREEN
  BLUE
}
```

It is here somewhat equivalent to Java `enum`, and can be extended through
augmentation.

### Option type

```golo
union Option = {
  Some = {value}
  None
}
```

Values can have fields. They are in this respect similar to `struct`

### Lists

```golo
union ConsList = {
  List = { head, tail }
  Empty
}
```

### Binary trees

```golo
union Tree = {
  Empty
  Leaf = { value }
  Node = { left, right }
}
```

## Augmentations

The `union` itself is a abstract class, and each possible value is a concrete
final (inner) class extending it. It is thus possible to augment the whole `union`:
```golo
augment Option {
  function map = |this, func| -> match {
    when this == Option.None() then this
    otherwise Option.Some(func(this: value()))
  }
}
```
or just a value:
```golo
augment ConsList$Empty {
  function isEmpty = |this| -> true
  function head = |this| -> null
  function tail = |this| -> this
}
augment ConsList$List {
  function isEmpty = |this| -> false
}
```

## Remark

I want to provide a neat namespace for the generated classes, to allow the
definition of unions with values having the same name in the same module.

Currently, the union itself is an abstract class and each value is an inner
class of the union. The FQN of a value type is thus `module.types.MyUnion$Value`.
The abstract class provides static factory methods to create the values, used 
as `MyUnion.Value(fields...)`

As an example, a `union` declaration like:
```golo
union Option = {
  Some = { value }
  None
}
```

generates classes equivalents to:
```java
abstract class Option {

  public static final class Some extends Option {
    public final Object value;
    protected Some(Object v) { this.value = v; }

    @Override
    public String toString() {
      return "enum Option.Some{value=" + value + "}";
    }

    @Override
    public boolean equals(Object other) {
      if (other == this) { return true; }
      if (other == null) { return false; }
      if (!(other instanceof Some)) { return false; }
      Some o = (Some) other;
      return java.util.Objects.equals(value, o.value);
    }

    @Override
    public int hashCode() {
      return java.util.Objects.hash(value);
    }
  }

  public static Option Some(Object value) {
    return new Some(value);
  }


  public static final class None extends Option {
    @Override
    public String toString() {
      return "enum Option.None";
    }
  }

  public static final Option None = new None();

}
```

An other solution is to use public classes in a sub-package, giving types like
`module.types.myunion.Value` and to generate factory functions in the defining 
module (as it is the case for `struct`)

I'm not sure what is the more convenient. What do you think?


yloiseau:
Comments on syntax and implementation are welcome

jponge:
Waoo...

I'll have a look tomorrow but it looks like you have done something
terrific :-)

yloiseau:
to be done:

[ ] tests... (I known, TDD next time)
[ ] golodoc integration

danielpetisme:
How do you do that.... every you touch is automatically converted into a mathematical concept ! :wink: 
Hats off

yloiseau:
@danielpetisme functional programming power! :v:

yloiseau:
BTW, I’m not sure about the keyword name, which could be misleading. Other possibilities could be:
- `union`
- `variant`
- `sum` (also misleading)
- `sumType`

Any toughts?

danielpetisme:
I vote for
* `enum` (my first thought is java enum but If I'm right is not excatly the same)
* `variant` (cause it's a new keyword in my Java world so it force me to open some docs to understand the concept)

Otherwise I could invent a term (I miss `pimp` :wink: )

jponge:
`enum` looks good to me. `union`, too.

doc/enums.asciidoc
```diff
@@ -0,0 +1,155 @@
...
+  let noString = Option.None()
+  println(noString)
+}
+----
+
+=== Usage example ===
+
+Some well known usages of sum types are:
+
+==== Enumerations
```
jponge: I think you can go for a consistent header style rather than sometimes `== foo ==` and sometimes `== foo` :smiley:

jponge: BTW `== title` is more idiomatic asciidoc.

samples/enum.golo
```diff
@@ -0,0 +1,109 @@
...
+  None
+}
+
+augment Option {
+  function bind = |this, func| -> match {
+    when this is Option.None() then this
+    otherwise func(this: value())
+  }
+
+  function fmap = |this, func| -> match {
```
jponge: `flatMap`

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeEnumGenerator.java
```diff
@@ -0,0 +1,267 @@
...
+    mv.visitInsn(IRETURN);
+    mv.visitLabel(sameType);
+
+    // cast other to <value>
+    mv.visitVarInsn(ALOAD, 1);
+    mv.visitTypeInsn(CHECKCAST, target);
+    mv.visitVarInsn(ASTORE, 2);
+
+    // java.util.Objects.equals(<member>, other.<member>)
+    for (String member : value.getMembers()) {
```
jponge: Isn't `Objects.equals(...)` already doing the above checks?

src/main/java/fr/insalyon/citi/golo/compiler/PackageAndClass.java
```diff
@@ -78,6 +85,14 @@ public String toJVMType() {
     return toString().replaceAll("\\.", "/");
   }
 
+  /**
+   * @return a JVM reference type representation for this object, e.g.: <code>foo.Bar</code> gives
```
jponge: I'd rather use `{@code }` blocks in Javadocs.

jponge:
That's a lovely pull request :clap:

danielpetisme:
Could it be a good practice to separate all the bytecode generation in dedicated classes? Just as you did w/ `JavaBytecodeEnumGenerator`.
Currently the class [JavaBytecodeGenerationGoloIrVisitor](https://github.com/golo-lang/golo-lang/blob/master/src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java) looks like the almighty God of Bytecode generation (up to 800 LoC).
Thoughts?

jponge:
It is not necessarily very easy to split `JavaBytecodeGenerationGoloIrVisitor`. We already have a class for structs bytecode generation.

It's big, but quite readable and methods aren't too fat :-)

samples/enum.golo
yloiseau: To match `java.util.Optional`, `bind` should be `flatMap` and `fmap` should be `map`. I used here the classical monad vocabulary, but I can switch to Java if you prefer

yloiseau:
@jponge thanks, it was fun to do... 
@danielpetisme I just mimic the `struct` implementation, without considering refactoring. As @jponge said, the split is not easy since the visitor keeps a global state if I recall correctly.

doc/enums.asciidoc
yloiseau: the current golo documentation mainly use `== title ==` :wink:

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeEnumGenerator.java
yloiseau: `Objects.equals` is used here to compare the members, while the checks above tests for the instance itself

src/main/java/fr/insalyon/citi/golo/compiler/PackageAndClass.java
yloiseau: copy-pasted from above :smile:

jponge: So that was also bad :-)

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeEnumGenerator.java
jponge: :+1:

doc/enums.asciidoc
jponge: Which is bad :-)

samples/enum.golo
jponge: I prefer flatMap

Artpej:
I prefer `union` to keep a certain consitency with `struct` and to be less confising with java `enums`.

yloiseau:
I think I'm gonna rename all stufs as `union`, to be more clear that it’s more than an equivalent of java `enum`

jponge:
I like `union` :+1:

yloiseau:
I think this PR is ready (since highlighting is elsewhere and a lib of useful unions can be done in an other PR).
 Any other comments?

doc/unions.asciidoc
```diff
@@ -0,0 +1,155 @@
+== Unions
+
+Golo allows the definition of sum algebraic data types, also known as
+_tagged_union_
+footnote:[ http://en.wikipedia.org/wiki/Tagged_union[Wikipedia: Tagged union]
```
jponge: Can't we just link instead of a footnote?

doc/unions.asciidoc
```diff
@@ -0,0 +1,155 @@
...
+  println(aString: value())
+
+  let noString = Option.None()
+  println(noString)
+}
+----
+
+=== Usage example
+
+Some well known usages of sum types are:
```
jponge: I'd suggest
> Some well known usages of sum types are the following.

doc/unions.asciidoc
```diff
@@ -0,0 +1,155 @@
...
+union Tree = {
+  Empty
+  Leaf = { value }
+  Node = { left, right }
+}
+----
+
+=== JVM existence
+
+An `union` type is compiled to an abstract JVM class. Each alternative value
```
jponge: A union type, not an :-)

jponge:
Minor comments on the documentation.

Otherwise this is a _huge_ feature, hats-off :smile:

jponge:
@yloiseau You can merge at your convenience.

doc/unions.asciidoc
yloiseau: I have used footnotes everywhere in the doc I wrote (augmentations, decorators)... I'll review that and change to simple links

PR #190, opened 2014-10-10
Change the original function arity with decorators

Artpej:
It is not currently possible to change the
the arity of a decorated function.
( see https://gist.github.com/yloiseau/d669c0e30023bf5b5bbc )

Originaly the function `foo`:

```golo
module curry

function curry = |f| -> |a| -> |b| -> f(a, b)

@curry
function foo = |a,b| -> a + b

```

is converted to :

```golo
module curry

function curry = |f| -> |a| -> |b| -> f(a, b)

function foo = |a,b| -> ^curry( |a, b| -> a + b )(a, b)
```

The trick tranforms the original function in a
variable arity function like this :

```golo
module curry

function curry = |f| -> |a| -> |b| -> f(a, b)

function foo = |args...| -> ^curry( |a, b| -> a + b ): invokeWithArguments(args)
```

PR #261, opened 2015-04-29
Boxing of MethodHandle into FunctionReference

jponge:
This is a backwards-incompatible change.
gololang.FunctionReference is a new class to box MethodHandle references.

The rationale for boxing function / closure references is that of attaching state to method handles,
which is not possible since MethodHandle is a final class, and is motivated by the future introduction
of named parameters.

At this stage the boxing does not add any new state. This will done in future works, but the nice
takeaway is that we now have an indirection in place.

Plugging FunctionReference in place of MethodHandle required lots of modifications. At this stage the
code passes all tests, but we will still need to:

1. test on larger code bases, and
2. port previous MH augmentations into the new FunctionReference class, and
3. update the documentation (user guide, {java,golo}docs).

Note that ClosureCallSupport still builds a method handle chain with inline-caching, hence performance
is not expected to suffer for regular calls such as:

	let f = |n| -> n + 1
	println(f(1))

Calls through methods are likely to remain as fast, since the unboxing indirection is likely to cause
handle() to be inlined:

	let f = |n| -> n + 1
	println(f: handle(): invoke(1))
    println(f: invoke(1))


src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -778,13 +783,21 @@ public void visitClosureReference(ClosureReference closureReference) {
           "java/lang/invoke/MethodHandles",
           "insertArguments",
           "(Ljava/lang/invoke/MethodHandle;I[Ljava/lang/Object;)Ljava/lang/invoke/MethodHandle;", false);
+      methodVisitor.visitTypeInsn(NEW, "gololang/FunctionReference");
+      methodVisitor.visitInsn(DUP_X1);
+      methodVisitor.visitInsn(SWAP);
+      methodVisitor.visitMethodInsn(
```
danielpetisme: Can't it be an indy invocation?

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -751,22 +751,27 @@ public void visitTryCatchFinally(TryCatchFinally tryCatchFinally) {
...
+          INVOKEVIRTUAL,
+          "gololang/FunctionReference",
+          "handle",
+          "()Ljava/lang/invoke/MethodHandle;", false);
       String[] refs = closureReference.getCapturedReferenceNames().toArray(new String[syntheticCount]);
       loadInteger(methodVisitor, 0);
       loadInteger(methodVisitor, syntheticCount);
       methodVisitor.visitTypeInsn(ANEWARRAY, "java/lang/Object");
+      ReferenceTable table = context.referenceTableStack.peek();
```
danielpetisme: why do we need to add a `peek`. I'm just trying to improve my bytecode skills not challenging the impl.

danielpetisme:
seems you had fun with _Find `MethodHandle` Replace with `FunctionReference`_
I must admit that my major issue is with the explicit `handle()` indirection but I can live with it.
Nice job!
Looking forward using it for named args. (and other freaky ideas)

src/main/java/fr/insalyon/citi/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
jponge: It could, but when you can avoid it it's simpler :-)

jponge:
@danielpetisme find & replace would have been simpler if I had been able to properly delegate the `invokeXYZ(...)` methods, but in such cases I got nasty array boxing issues. This is why I took the decision to introduce `handle()`.

jponge:
This pull-request is ok from my side.

Artpej:
Seem's all right to me too, I will adapt the new decorators implementation (#190) when this one will be merged. :+1:

Issue #262, opened 2015-05-06
Improve function chaining error messages


danielpetisme:
Considering this snippet
```
module chaining


function main = |args...| {
  let f = |x, y| -> x + y
  let g = |a, b| -> a + b
  let h = f: andThen(g)

  h: invoke(1, 2)
}
```
Obviously the function chain can' t work as is because `g` is expecting 2 args when `f` will return one and only one value. 

The according error message (Which frankly freaked me out!) is:
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000144d8ec62, pid=12674, tid=4867
#
# JRE version: Java(TM) SE Runtime Environment (8.0_45-b14) (build 1.8.0_45-b14)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.45-b02 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# j  java.lang.invoke.DirectMethodHandle.internalMemberName(Ljava/lang/Object;)Ljava/lang/Object;+0
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/daniel/workspace/golo-temp/hs_err_pid12674.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
[1]    12674 abort      golo golo --files chaining.golo
```

From a user point of view is 1/ scary to see a message raise by the JVM itself 2/ helpless to find  and fix the root problem.

Would be nice to detect such mistakes and warn the developer.

If I'm right the filter function (ie. the `andThen` argument) can have 0 or 1 argument (no more). So we could easily detect that during the invocation of the brand new `FunctionReference`.

yloiseau:
the composing functions (this is a function composition, not a filter :p) must have 1 parameter, since they are called with the result of the previous function as argument. This could be checked indeed. 
We could even go farther, by autospreading arguments, such that:
```golo
let f = |x, y| -> [x + y, x - y]
let g = |a, b| -> a + b
let h = f: andThen(g)
require(h(21) == 42, "err")
```
would work; kinda implicit uncurry, but explicit is better than implicit.

jponge:
Oh yes we could add a check in andThen()

danielpetisme:
[...]

PR #263, opened 2015-05-07
Check andThen function parameter arity

jponge:
Checks that the function argument takes exactly 1 parameter.
Interestingly, the composition of method handles with filterReturnValue and mismatching
types is possible, and provokes a SIGSEGV of the JVM at runtime.

Fix for #262

Signed-off-by: Julien Ponge <julien.ponge@insa-lyon.fr>


danielpetisme:
Seems nice,  :+1:

jponge:
Yeah, stupid enough to do the job :smiley: 

Merging!

Issue #268, opened 2015-05-25
Boxing problem appearing on augmented `List#join` when all tests ran


danielpetisme:
Teaser: This is a tricky one

In the standard-augmentations, the current impl of `List#join` (mine in fact :confused:) raise an exception when all tests are ran but not if your run the failing-ish tests.

*how to reproduce*
Add the following method in any golo code used for tests
```
function plop = -> list[1, 2, 3]: join("-")
```
Add the associated Java code
```
Method plop = moduleClass.getMethod("plop");
assertThat((String) plop.invoke(null), is("1-2-3"));
```

Now run `mvn clean test`, 2 tests fail:
* `DocumentationRenderingTest.html_processor`
* `DocumentationRenderingTest.markdown_processor`
```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
        at sun.invoke.util.ValueConversions.primitiveConversion(ValueConversions.java:242)
        at sun.invoke.util.ValueConversions.unboxInteger(ValueConversions.java:76)
        at gololang.StandardAugmentations$java$util$List.join(standard-augmentations.golo:430)
        at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:625)
        at fr.insalyon.citi.golo.runtime.MethodInvocationSupport.fallback(MethodInvocationSupport.java:191)
        at anonymous317906307738923.__$$_closure_0($Anonymous$_317906307759624.golo:137)
        at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:625)
        at fr.insalyon.citi.golo.doc.MarkdownProcessor.render(MarkdownProcessor.java:34)
        at fr.insalyon.citi.golo.doc.DocumentationRenderingTest.markdown_processor(DocumentationRenderingTest.java:53)
```
if you run unitary the test `mvn -Dtest=DocumentationRenderingTest#html_processor clean test` it passes.

Fix:
In fact I know how to fix the issue. It comes from the current `join` implementation
```
1  function join = |this, separator| {
2     var buffer = java.lang.StringBuilder("")
3     if not (this: isEmpty()) {
4       buffer: append(this: head())
5       let tail = this: tail()
6       if not (tail: isEmpty()) {
7         buffer: append(separator)
8         buffer: append(tail: join(separator))
9       }
10    }
11   return buffer: toString()
12  }
```

Replacing the line`#4` by `buffer: append(this: head(): toString())` (adding a `toString())` fix the issue.

[...]

PR #269, opened 2015-05-26
Refactoring the augmented join method

danielpetisme:
Since `Tuple` and `List` have both a `get` method the `join` algo can be extracted and mutualized. 
This is a quick, naive and dirty workaround fo #268 .


jponge:
:+1:

PR #272, opened 2015-05-29
Allows `fun` to retrieve non-static methods.

yloiseau:
This allows `gololang.Predefined::fun` to retrieve non-static methods,
and so to access unbound methods. It is now possible to write:
```golo
list["Foo", "Bar", "Hello", "Goodbye"]
  : filter(^String::endsWith: bindAt(1, "o"))
```
instead of
```golo
list["Foo", "Bar", "Hello", "Goodbye"]
  : filter(|s| -> s: endsWith("o"))
```
or
```golo
list["Foo", "Bar", "Hello", "Goodbye"]
  : map(^String::length)
```

Also fixes a NPE on named parameters in `FunctionReference`.

### DIFF
--- a/doc/closures.asciidoc
+++ b/doc/closures.asciidoc
@@ -269,6 +269,22 @@ function call_local_fun = {
 }
 ----
 
+Note that you can in the same way get a reference to a Java static method. For
+instance:
+
+[source,golo]
+----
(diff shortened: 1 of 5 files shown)


jponge:
Go for the merge! :smile:

PR #274, opened 2015-06-03
Add a fallback method support for augmented classes

danielpetisme:
Tired to catch `NoSuchMethodError`s ? This PR aims to propose a fallback behavior to intercept failed method dispatch. 

Fallbacks permits to implement dynamic methods, DSLs  or Fluent API

*Named Argument + Fluent example*

```golo
augmentation Fluent = {
  function fallback = |this, name, args...| {
    return match {
      when args: length() is 1 then this: bindAt(name, args: get(0))
      otherwise this: bindAt(name, args)
    }
  }
}
augment gololang.FunctionReference with Fluent

function greet = |firstName, lastName| -> "Hello " + firstName + " " + lastName + "!"

function main = |args...| {
  #Partial application based on the argument name rather than the argument position
  let msg = ^greet: lastName("Doe"): firstName("John"): invoke()
  println(msg)
  #Prints "Hello John Doe!"
}
```
Thoughts?


src/test/resources/for-execution/augmentations-with-fallback.golo
```diff
@@ -0,0 +1,34 @@
+module golotest.execution.AugmentationsWithFallback
+
+import java.util.stream.Collectors
+
+augment java.lang.String {
+  function fallback = |this, name, args...| -> "Fallback for this " + this + " named " + name + " with args" + args: asList()
+}
+
+augment Fluent {
```
yloiseau: augmentation Fluent = {

yloiseau:
:+1:

src/test/resources/for-execution/augmentations-with-fallback.golo
danielpetisme: the code above works (w/ `augment Fluent`)... is it normal?

yloiseau: Yes it's a side effect of the fact that all is in the same module.

danielpetisme: is it documented? I' ll change the tests and the snippet

yloiseau: No it's not. It just a "bug" in the names of the generated classes that are the same in this precise case.

jponge:
Sounds cool.

jponge:
Is it Yannick-proof?

danielpetisme:
`fallback` is an _interceptor_ for `NoSuchMethodError`. Is it meaningfull to think about a larger _interceptor_ mechanism ?

jponge:
> fallback is an interceptor for NoSuchMethodError. Is it meaningfull to think about a larger interceptor mechanism ?

I see why you would want that, but can you elaborate?

danielpetisme:
The aim is to intercept:
* Method calls (exact match and/or patterns `find{Any,All,One}`
* Exceptions in a generalized way: I want to catch all the `IllegalArgumentException` my POJO can raise.

Decorators permits a micro-approach. Their limits are:
* I can only decorate existing method
* it requires an explicit instruction(`@decorate`) which make impossible to decorate exisiting 3rd party methods (I can'tdecorate the JDK for instance.
* I have to decorate every single function I have to

Some use cases
1. The easy one, I want to mimic the `missingProperty` method to permit Dynamic methods (grails Dynamic finders for instance. (basically what I try to do w/ the fallback PR)
```
augment java.util.List {
  function interceptor = |this| -> {
    on(NoSuchMethodError.class): invoke(^fallback)
  } 
}
```

* I want to transform actual Java synchronous method calls into async ones wiithout impacting my program API 
```
augment java.util.List {
  function interceptor = |this| -> {
    on("add"): invoke(^asyncAdd)
  } 
}
```

* I want a to benchmark every method calls
```
augment java.util.List {
  function interceptor = |this| -> {
    on("*"): invoke(^myBenchDecorator)
  } 
}
```

Thoughts?

jponge:
Sounds interesting.

Ceylon has interceptors: http://staging.ceylon-lang.org/documentation/1.0/tour/interceptors/

I'd go with a merge of that PR, and possibly evolve a more elaborated interceptors support in a following one.

danielpetisme:
I didn't know that but it's exactly the spirit interacting with runtime events (invocations, exceptions, etc).

jponge:
Yes, they have an interesting implementation.

danielpetisme:
I'm interrested about feedbacks about the implementation. 
In particular 
* https://github.com/danielpetisme/golo-lang/blob/wip/augmentation-fallback/src/main/java/fr/insalyon/citi/golo/runtime/MethodInvocationSupport.java 
I _hardcoded_ the `fallback` callsite (including the argument names).
It works but I think it can be cleaner

src/main/java/fr/insalyon/citi/golo/runtime/MethodInvocationSupport.java
```diff
@@ -159,6 +160,26 @@ public static Object fallback(InlineCache inlineCache, Object[] args) throws Thr
     Class<?> receiverClass = args[0].getClass();
     MethodHandle target = lookupTarget(receiverClass, inlineCache, args);
 
+    if (target == null) {
```
yloiseau: I'd rather implement this aspect as a new strategy in `findTarget`

yloiseau: e.g. `InterseptionMethodFinder`, will be easier for the next step :wink:

danielpetisme: My constraints:
* the `fallback` resolution is made only when the first resolution failed.
  * Easy to handle in a strategy
* if the `fallback` has to be invoked then you have to create a specific callsite 
  * `findTarget` return `MethodHandle` this is the tricky part
* If the `fallback` resolution failed then the displayed message should only mention the initial method invoked (not the fallback one)
  * Easy

danielpetisme: After a first view, refactoring the snippet isn't easy...

jponge:
Shall we merge this PR now, or shall we wait for further developments?

danielpetisme:
for me it's _feature complete_. We tried with @yloiseau to refactor a bit the code (very quickly) and it doesn't as obvious as we thought.
Do you prefer to merge it and then we try to refactor the whole Finding strategy or do you want this PR to carry the refactoring?

jponge:
I'd go for merging it now, and you'll get back to the refactoring later on when at our new home :-)

jponge:
So i'm going to merge it.

jponge:
Thanks!

Issue #273, opened 2015-06-03
Exception when adapting a golo closure to a Java8 functional interface


yloiseau:
When we want to use a golo closure as a Java8 functional interface that returns a *primitive type*, the adaptation fails, and we got a `java.lang.AbstractMethodError`.

For instance, the given Golo code:
```golo
let f = Foo(2)
f: bar(|v| -> v)
```
works with `Foo` declared as
```java
public class Foo {
  private int v;
  public Foo(int v) {
    this.v = v;
  }

  public Integer bar(Function<Integer,Integer> f) {
    return f.apply(v);
  }
}
```

but if
```java
public class Foo {
  private int v;
  public Foo(int v) {
    this.v = v;
  }

  public int bar(IntUnaryOperator f) {
    return f.applyAsInt(v);
  }
}
```
golo raises 

```
Exception in thread "main" java.lang.AbstractMethodError: Method pb$$Lambda$1.applyAsInt(I)I is abstract
```

This issue also raises with `Stream::filter` that accept a `Predicate` which returns a `boolean`

yloiseau:
More investigation on this:

given the Java code:
```java
public class Foo {
  public static boolean foo(Object v) {
    return true;
  }

  public static boolean bar(Predicate<Object> f) {
    return f.test(0);
  }
}
```

the following Golo code gives `true` as expected:
```golo
Foo.bar(^Foo::foo)
```

Changing the Java into
```java
public static Boolean foo(Object v)
```
the same Golo code raises 
```
AbstractMethodError: Method pb$$Lambda$1.test(Ljava/lang/Object;)Z is abstract
```

My interpretation:
in `runtime.FunctionCallSupport`, the `metafactory` creates an instance of `Predicate<Object>`, binding `foo` to `test(Object)Boolean` and thus the interface method `test(Object)boolean` is (somewhat) *overloaded* instead of *overriden*, thus the `AbstractMethodError`.

yloiseau:
Adding a `.asType(lambdaType)` or `MethodHandles.explicitCastArguments` also fails. The types are correct, but the `metafactory` raises a `IllegalArgumentException: not a direct method handle`, indeed the handle is a `java.lang.invoke.BoundMethodHandle$Species_LL`... (may be linked to, or similar to, http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8020816)

PR #276, opened 2015-06-04
Fix bug #273.

yloiseau:
This PR fixes bug #273.

We had a problem when adapting a Golo closure to a Java8 lambda, when
the required functional interface returns a primitive type. This
appears for instance when using `Stream::filter`, since
`Predicate::test` returns a `boolean`.
Fixed by forcing the implemented signature in the `metafactory` to the required one.


jponge:
:+1: :smile:

PR #280, opened 2015-06-11
Lazy lists

yloiseau:
This PR adds lazy lists to Golo.

Lazy lists are lists in which the tail is actually a closure that is evaluated only if needed. Thus, it is possible to create infinite lists, and map functions on these infinite lists. By definitions, a lazy list is immutable.

The provided structure is a `HeadTail` and a `Collection`, and can transparently be used in classical recursive functions with constructs like
```golo
let sum = |lst| -> match {
  when lst: isEmpty() then 0
  otherwise lst: head() +  sum(lst: tail())
}
```
as well as iteration constructs such as
```golo
let sum = |lst| {
  var s = 0
  foreach elt in lst {
    s += elt
  }
  return s
}

### DIFF
--- a/samples/lazylist.golo
+++ b/samples/lazylist.golo
@@ -1,0 +1,55 @@
+# ............................................................................................... #
+#
+# Copyright (c) 2012-2015 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
+#
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# ............................................................................................... #
+
+module samples.LazyLists
+
+import gololang.lazylist
+
+----
+create a lazy list of all integers
+----
+function count = |start| -> cons(start, -> cons(start + 1))
+
+function useCount = {
+
+  # the list of all even multiples of 3
+  let l = count(): map(|x| -> 3 * x): filter(|x| -> (x % 2) == 0)
+  # nothing is computed yet
+
+  foreach v in l: take(10) {
+    println(v)  # print the 10 first even multiples of 3
+  }
+}
+
+
+----
+create a lazy range of integers
+----
+function lazyRange = |start, end| -> match {
+  when start >= end then emptyList()
+  otherwise cons(start, -> lazyRange(start + 1, end))
+}
+
+----
+reimplementation of the map function
+----
+function myMap = |lst, func| -> match {
+  when lst: isEmpty() then emptyList()
+  otherwise cons(func(lst: head()), -> myMap(lst: tail(), func)))
+}
+
+# a lazy list of known values
+let lst1 = cons(1, cons(2, cons(3, cons(4, emptyList()))))
+
+# or simpler
+let lst2 = lazyList(1, 2, 3, 4)
+
+
--- a/src/main/golo/lazylist.golo
+++ b/src/main/golo/lazylist.golo
@@ -73,13 +73,13 @@ function cons = |head, tail| -> match {
   otherwise gololang.LazyList(head, -> tail)
 }
 
-----
-Unary version of [`cons(head, tail)`](#cons_head_tail).
-
-Its parameter is assumed to be a tuple (or any object having a `get(idx)` method)
-of the form `[head, tail]`.
-----
-function cons = |ht| -> cons(ht: get(0), ht: get(1))
+#----
+#Unary version of [`cons(head, tail)`](#cons_head_tail).
+#
+#Its parameter is assumed to be a tuple (or any object having a `get(idx)` method)
+#of the form `[head, tail]`.
+#----
+#function cons = |ht| -> cons(ht: get(0), ht: get(1))
 
 ----
 Variadic function to create lazy lists from values.
@@ -220,7 +220,7 @@ augment gololang.LazyList {
   function dropWhile = |this, pred| -> match {
     when this: isEmpty() then Empty()
     when not pred(this: head()) then this
-    otherwise this: tail(): drop(pred)
+    otherwise this: tail(): dropWhile(pred)
   }
   ----
   Filters elements based on a predicate.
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by yloiseau** [comment]
I don't know or I keep the hybrid aspect of the code or if I move the golo functionalities directly in the java class.

**Comment by jponge** [comment]
Looks bright!

File: src/main/java/gololang/LazyList.java [hunk 1]
```diff
@@ -0,0 +1,396 @@
...
+   *
+   * @return a {@code LazyList}, or {@code EMPTY} if the list is empty,
+   * contains only one value, or if the closure failed.
+   */
+  public LazyList tail() {
+    if (memoTail == null) {
+      try {
+        memoTail = (LazyList) (this.tail.invoke());
+      } catch (Throwable e) {
+        memoTail = EMPTY;
```
MakeNowJust: Why is it ignore an exceptions?  Ignoring exception is the popular cause of bugs, so it should rethrow an exception.

However, what dose it perform when second calling of `tail` method?

yloiseau: There are many cases where this exception can't be properly handled by the user (e.g. when performing a `map` on the list). In this case, it seems more practical to simply terminate the list. This is semantically correct w.r.t. the lazyness of the list. If the next function can't perform, then the list is empty. The `tail` closure is never called twice, since the result is cached. Since the functions representing the next value *must* be pure functions with no parameters, it *must* always return the same value (or fail to run each time).

MakeNowJust: @yloiseau It makes sense. It's right. Perhaps, I was  misled use case of `LazyList` (I thought it looks like `Stream`). Thank you.

yloiseau: Actually, a lazy list is quite similar to a java `Stream`.

jponge:
Is this complete?

src/main/java/gololang/HeadTailIterator.java
```diff
@@ -0,0 +1,32 @@
+package gololang;
```
jponge: Missing header file

yloiseau:
More function for specific generators (e.g. `count(x) -> x, x+1, x+2, ...`) and augmentations on lazy list (specially for infinite ones, e.g. `take(n)`, `takeWhile(cond)`, `drop(n)`, `dropWhile(cond)`, etc.) could be added, but it can take place in another PR, at your convenience.

jponge:
Just let me know what you prefer Yannick.

Either we merge now and you complete it through another PR after the Eclipse move, or we finish it in a single PR.

yloiseau:
Depends on the deadline. BTW, should I add a documentation section or is the javadoc/golodoc enough?

jponge:
The freeze deadline is this Friday, but we could jump in earlier for the initial contribution if all pending PRs are sorted out.

A small piece in the documentation would help, e.g. http://golo-lang.org/documentation/next/#_misc_modules

yloiseau:
Ok, so I'll fix some typos, review comments, write a small doc section, and we can merge.

jponge:
:+1: 

BTW if you want to anticipate and use EPL license headers... :-) (see the other branch)

yloiseau:
That's what I done for the missing one. I'll change the other ones

yloiseau:
If you are ok with the code, I think we can merge.

samples/lazylist.golo
```diff
@@ -0,0 +1,55 @@
+# ............................................................................................... #
```
jponge: I'll fix that one: for samples we should rather use the Eclipse Distribution License (it's a BSD, really)

src/main/golo/lazylist.golo
```diff
@@ -0,0 +1,355 @@
+# ............................................................................................... #
+#
+# Copyright (c) 2012-2015 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
```
jponge: Note that once we are at Eclipse you will no longer be required to assign all copyright to INSA, since you will be a committer at Eclipse, and you will be just fine with your own copyright mention :-)

jponge:
I'm going to merge it, but first I'm merging the Eclipse preparation branch.

jponge:
Ok, there is an issue :smile: 

You have both `gololang.LazyList` (a Java class) and `gololang.lazylist` (a Golo module).

1. This is an issue on case-insensitive filesystems (HFS on MacOSX is like that by default, I'd bet for NTFS too).
2. This triggers an error with the OSGi generation plugin:

```
INFO] --- maven-bundle-plugin:2.5.0:manifest (bundle-manifest) @ golo ---
[ERROR] Manifest org.golo-lang:golo:jar:3.0.0-incubation-M1-SNAPSHOT : Classes found in the wrong directory: {gololang/LazyList.class=gololang.lazylist}
[ERROR] Error(s) found in manifest configuration
```

How would you like to fix it?

BTW if you want to edit `samples/lazylist.golo`, then the header shall be like:

```
# Copyright (c) 2012-2015 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
#
# All rights reserved. This Example Content is intended to demonstrate
# usage of Eclipse technology. It is provided to you under the terms and
# conditions of the Eclipse Distribution License v1.0 which is available
# at http://www.eclipse.org/org/documents/edl-v10.php
```

yloiseau:
Ok. Maybe renaming the golo module to `gololang.lazylists`? It makes sens since it's a bunch of utility functions and augmentations.

jponge:
Sounds good. And name it `gololang.LazyLists` for consistency?

yloiseau:
other golo modules in `gololang` are lowercase...

jponge:
Oh damn... then you can keep it lowercase

jponge:
https://github.com/golo-lang/golo-lang/tree/master/src/main/golo

No they are in camel case :-)

yloiseau:
like `standard-augmentations`? :smile:
I think we already had this discussion, but maybe we should define some rules and conventions to keep things consistents.

jponge:
I'm talking about the module name, not the file name :-)

yloiseau:
Damn... ok, sorry

jponge:
Merged!

PR #287, opened 2015-06-30
[471130] Guarded loops

jponge:
This introduces support for guards in foreach loops, as in:

```golo
foreach arg in args when arg: length() > 2 {
  println("  " + arg)
}
```

This is syntaxic sugar, equivalent to:

```golo
foreach arg in args {
  if arg: length() > 2 {
    println("  " + arg)
  }
}
```


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471130

PR #288, opened 2015-07-02
[471709] Exception when comparing numbers

yloiseau:
The implementation of comparison operators was done using `compareTo`,
which fails when comparing two numbers of different classes.

The bug is fixed by using `Double.compare` on the two numbers
`doubleValue` when both values are instances of `Number`, falling back
to `compareTo` otherwise.

The same is done for equality, since until now, golo considered
`1 != 1.0`. When the two values are numbers, the double value is
compared by value, `equals` is called otherwise.

Bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=471709

### DIFF
--- a/src/test/resources/for-execution/operators.golo
+++ b/src/test/resources/for-execution/operators.golo
@@ -82,44 +82,43 @@ function polymorphic_number_comparison = {
   let left = list[1, 1_L, 1.0, 1.0_F]
   let right = list[2, 2_L, 1.1, 1.1_F]
 
-  var res = true
   foreach a in left {
     foreach b in right {
-      res = res and (not (a == b))
-      res = res and (not (b == a))
-      res = res and (b != a)
-      res = res and (a != b)
+      require((not (a == b)), "equals failed")
+      require((not (b == a)), "equals failed")
+      require((b != a), "not equals failed")
+      require((a != b), "not equals failed")
       
-      res = res and (a <= b)
-      res = res and (not (a >= b))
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=471709

jponge:
There is a better way :smile: 

https://github.com/eclipse/golo-lang/blob/master/src/main/ruby/

(Just run them copy and paste the text in OperatorSupport)

If you add generation of comparison operators then you can have type-specific targets, and the binary operator inline-cache can do the rest, just like with the arithmetic operators. And that can be fast / inline, at least faster than a more generic target as you have now.

src/test/resources/for-execution/operators.golo
```diff
@@ -77,3 +77,49 @@ function null_guarded = {
   let map = map[]
   return map: get("bogus") orIfNull "n/a"
 }
+
+function polymorphic_number_comparison = {
```
jponge: Perhaps the checks could be enforced with a `raise` rather than just returning `res`?

yloiseau: you mean a series of `require(a==b, "equals failed")` with no check in the java unit test?

jponge: > you mean a series of require(a==b, "equals failed") with no check in the java unit test?

Yes, + the unit test of course.

PR #289, opened 2015-07-04
Feature/simple destruct

yloiseau:
Allows to destructure an object to assign to several values at the same
time. E.g.:

```golo
let a, b = [1, 2]
require(a == 1 and b == 2, "failed")
```

Any object with a `destruct` method returning a tuple can be used.
Vararg can be used as the last value, e.g.

```golo
let a, b... = [1, 2, 3, 4]
require(a == 1 and b == [2, 3, 4], "failed")
```

Also allows constructs like
```golo
foreach key, value in map: entrySet() {
  println(key)
  println(value)
}
```
on any iterable, as long as values are destructurable.

### DIFF
--- a/doc/basics.asciidoc
+++ b/doc/basics.asciidoc
@@ -453,6 +453,14 @@ let lst = list[1, 2, 3, 4, 5]
 let head, tail... = lst        # head = 1, tail = [2, 3, 4, 5]
 ----
 
+Already defined variables can also be assigned with destructuring. For
+instance, one can easily swap two variables:
+[source,golo]
+----
+var a, b = [1, 2]  # a = 1, b = 2
+a, b = [b, a]      # a = 2, b = 1
+----
+
 Destucturing can also be used in
 xref:_span_class_monospaced_foreach_span_loops[`foreach` loops]:
 [source,golo]
--- a/doc/basics.asciidoc
+++ b/doc/basics.asciidoc
@@ -409,6 +409,61 @@ let b = -> [2, 'b']
 let m = map[a(), b()]
 ----
 
+=== Destructuring ===
+
+Golo supports simple destructuring, that is automatic extraction of values
+from an object and assignment to multiple variables in one instruction.
+
+For instance, using destructuring on a tuple:
+[source,golo]
+----
+let a, b = [1, 2]
+# a = 1, b = 2
+----
+
+If there are more variables than values, an exception is raised. If there are
+fewer, the remaining values are ignored. A special syntax is available to
+assign the rest of the values, similar to varargs notation. For instance:
+
+[source,golo]
+----
+let a, b, c = [1, 2]               # raises an exception
+let a, b = [1, 2, 3]               # a = 1, b = 2, 3 ignored
+let a, b, c... = [1, 2, 3, 4, 5]   # a = 1, b = 2, c = [3, 4, 5]
+----
+
+Any object having a `destruct()` method returning a tuple can be used in
+destructuring assignments. Golo specific data structures and some Java native
+ones can be destructured. Augmentations can be used to make an existing class
+destructurable.
+
+For instance, xref:_structs[golo structures] are destructurable:
+[source,golo]
+----
+struct Point = {x, y}
+#...
+let p = Point(42, 1337)
+let x, y = p   # x = 42, y = 1337
+----
+
+as well as java lists:
+[source,golo]
+----
+let lst = list[1, 2, 3, 4, 5]
+let head, tail... = lst        # head = 1, tail = [2, 3, 4, 5]
+----
+
+Destucturing can also be used in
+xref:_span_class_monospaced_foreach_span_loops[`foreach` loops]:
+[source,golo]
+----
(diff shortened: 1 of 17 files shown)


jponge:
And you can have the following, right?

```golo
let a, b, c = [1, 2, 3]
```

yloiseau:
of course... as stated in the doc:
```golo
let a, b, c = [1, 2, 3] # a = 1, b = 2, c = 3
let a, b = [1, 2, 3] # a = 1, b = 2, 3 ignored
let a, b, c = [1, 2, 3] # exception
let a, b, c... = [1, 2, 3, 4, 5] # a= 1, b = 2, c = [3, 4, 5]
```

yloiseau:
on the other hand, something like:
```golo
var a = 1
var b = 2
a, b = [b, a]
```
is not allowed. In the grammar, the destructuration must start with `let` or `var`. Maybe I should change that.

yloiseau:
Now two variables can be swaped 
```golo
var a = 1
var b = 2
a, b = [b, a]
```

jponge:
Lovely feature

yloiseau:
Thanks... it was mandatory since now even JavaScript has it :wink:

doc/basics.asciidoc
```diff
@@ -409,6 +409,69 @@ let b = -> [2, 'b']
...
+For instance, xref:_structs[golo structures] are destructurable:
+[source,golo]
+----
+struct Point = {x, y}
+#...
+let p = Point(42, 1337)
+let x, y = p   # x = 42, y = 1337
+----
+
+as well as java lists:
```
jponge: Perhaps you should add an example on maps, that is really useful.

jponge: Ha... nevermind, you have it in the `foreach` loop.

src/main/golo/standard-augmentations.golo
```diff
@@ -284,6 +284,12 @@ augment java.util.Collection {
   ----
   function newWithSameType = |this| -> _newWithSameType(this)
 
+  ----
+  Destructuration helper.
+  
+  * return a tuple of the values
+  ----
+  function destruct = |this| -> Tuple.fromArray(this: toArray())
```
jponge: We should maybe introduce a `copy()` method in `Tuple`.

src/main/java/fr/insalyon/citi/golo/compiler/parser/ASTDestructuringAssignment.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2012-2015 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
```
jponge: BTW as a committer you may assign another copyright notice to yourself, too.

src/main/java/gololang/AbstractRange.java
```diff
@@ -130,4 +131,12 @@ public T head() {
   public boolean isEmpty() {
     return from.compareTo(to) >= 0;
   }
+
+  public Tuple destruct() {
```
jponge: Why not making an array right from the start, and avoid a `LinkedList`?

src/main/java/gololang/Tuple.java
```diff
@@ -182,10 +182,22 @@ public Object head() {
    */
   @Override
   public Tuple tail() {
+    return this.subTuple(1);
+  }
+
+  /**
+   *
```
jponge: No doc? :-)

jponge:
Feel free to push it once you are confident with the implementation.

jponge:
@yloiseau to match JS we'd need the `===` operator I guess :laughing:

rlespinasse:
Nice indeed ... for the `simple destruct` feature :smile:

src/main/golo/standard-augmentations.golo
yloiseau: We are here augmenting `Collection`. Or did you mean `Tuple copy(Collection<Object>)` ?

jponge: > We are here augmenting Collection. Or did you mean Tuple copy(Collection<Object>) ?

I was thinking of something like `t: copy()`

yloiseau: We can but it wouldn't be usable here (`this` is a collection...)

yloiseau: And since tuples are immutables, I don't realy see the need for it anyway

jponge: Ok, just forget :-)

yloiseau:
BTW I was thinking... I don't know why I started with tuples (maybe to deal with “real” objects), but wouldn't be simpler if `destruct` returned an array instead of a tuple?

jponge:
> BTW I was thinking... I don't know why I started with tuples (maybe to deal with “real” objects), but wouldn't be simpler if destruct returned an array instead of a tuple?

Tuples are fine with me.

yloiseau:
Ok, let's keep tuples then

PR #290, opened 2015-07-05
Fix special characters escaping in String and char

Artpej:
- fixes this:
   ```golo
   print("<\\t>")
   ```
   will print:
   ```
   <\    >
   ```
   but the expected value is
   ```
   <\t>
   ```

- add the ability to the parser to deal with special characters using single quote notation


jponge:
:+1: thanks @Artpej

PR #297, opened 2015-08-12
Adds special methods to unions.

yloiseau:
A union has now special methods to test its kind (type) as well as the
values of its members if applicable. This make the use of unions in
`match` constructs easier and more readable.

For instance, given the union:
```golo
union Tree = {
  Empty
  Leaf = { value }
  Node = { left, right }
}
```

one can now write a match like:
```golo
function foo = |elt| -> match {
  when elt: isEmpty() then "empty"
  when elt: isLeaf() then "leaf"
  when elt: isNode(Empty(), Empty()) then "node with empty children"
  when elt: isNode() then "a node"
}
```

In the same way, a union:
```golo
union Option = {
  Some = {value}
  None
}
```

can be tested with:
```golo
if v: isNone() {
  // ...
}
if v: isSome(null) {
  // a some, but containing null (not a good idea)
}
if v: isSome(42) {
  // a some containing 42
}
if v: isSome() {
  // a some, whatever le value
}
```

A special singleton class `Unknown` is added. The instance is equal to
any object (except null). It can be used to ignore members is the tests:
```golo
let _ = Unknown.get()
function getTreeValue = |elt| -> match {
  when elt: isEmpty() then null
  when elt: isLeaf() then elt: value()
  when elt: isNode(_, Empty()) then elt: left()
  when elt: isNode(Empty(), _) then elt: right()
  when elt: isNode() then [elt: left(), elt: right()]
  otherwise null
}
```

### DIFF
--- a/src/main/java/gololang/Unknown.java
+++ b/src/main/java/gololang/Unknown.java
@@ -5,8 +5,10 @@ package gololang;
  * A singleton class that is equals to any object.
  * <p>
  * Used in unions matching special methods to ignore a value.
+ * Note that this break the commutativity property of {@code equals}, and thus this object must be
+ * used with caution.
  */
-public class Unknown {
+public final class Unknown {
   private static final Unknown instance = new Unknown();
   private Unknown() {}
   public static Unknown get() { 
--- a/src/main/java/gololang/Unknown.java
+++ b/src/main/java/gololang/Unknown.java
@@ -1,0 +1,25 @@
+
+package gololang;
+
+/**
+ * A singleton class that is equals to any object.
+ * <p>
+ * Used in unions matching special methods to ignore a value.
+ */
+public class Unknown {
+  private static final Unknown instance = new Unknown();
+  private Unknown() {}
+  public static Unknown get() { 
+    return instance;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    return o == null ? false : true;
+  }
+
+  @Override
+  public int hashCode() { 
+    return 0; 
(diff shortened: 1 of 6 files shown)


src/main/java/gololang/Unknown.java
```diff
@@ -0,0 +1,25 @@
+
+package gololang;
+
+/**
+ * A singleton class that is equals to any object.
+ * <p>
+ * Used in unions matching special methods to ignore a value.
+ */
+public class Unknown {
```
jponge: How about making it a `final` class?

jponge: Ok, this is being picky, but still... :smiley:

src/test/resources/for-execution/unions.golo
```diff
@@ -178,6 +178,37 @@ function test_singleton = {
   }
 }
 
+function test_match_methods = {
```
jponge: Which makes me thing that we'd benefit from having a cool Golo-esque testing module.

jponge:
I only have a minor comment, and the documentation should be updated as well.

Other than that it's a very welcome addition :+1:

jponge:
This also makes me think that we don't have a `samples/` showcase for list comprehensions.

src/test/resources/for-execution/unions.golo
yloiseau: Indeed...

jponge:
Looks good to me.

PR #299, opened 2015-08-14
Project creation fix

jponge:
Proposed by a contributor on my own fork.

This ensures that nested folders are being properly created.


jponge:
Nevermimd

PR #301, opened 2015-08-24
Adds a ref predefined function creating reference

yloiseau:
The new predefined `ref` function returns an `AtomicReference` on the
given object. This is just a helper function around
`java.util.concurrent.atomic.AtomicReference`.

Useful whenever an explicit reference is needed, for instance to create
a closure with internal mutable state:

```golo
    function counter = |init| {
      let current = ref(init)
      return -> current: getAndSet(current: get() + 1)
    }

    function counter = |init, stepFun| {
      let current = ref(init)
      return -> current: getAndUpdate(stepFun)
    }

    let c = counter(3)
    c() # -> 3
    c() # -> 4
    c() # -> 5

    let c2 = counter(3, |x| -> x * 2)
    c() # -> 3
    c() # -> 6
    c() # -> 12
```

### DIFF
--- a/doc/predefined-functions.asciidoc
+++ b/doc/predefined-functions.asciidoc
@@ -299,7 +299,7 @@ println(e: getKey() + " => " + e: getValue())
 ----
 
 `box` gives instances of `java.util.concurrent.atomic.AtomicReference`, and can
-be used to create a muttable reference where not possible othewise, e.g. in
+be used to create a mutable reference where not possible otherwise, e.g. in
 closures, as in:
 
 [source,golo]
--- a/doc/predefined-functions.asciidoc
+++ b/doc/predefined-functions.asciidoc
@@ -297,3 +297,20 @@ let e = mapEntry("foo", "bar")
 # prints "foo => bar"
 println(e: getKey() + " => " + e: getValue())
 ----
+
+`box` gives instances of `java.util.concurrent.atomic.AtomicReference`, and can
+be used to create a muttable reference where not possible othewise, e.g. in
+closures, as in:
+
+[source,golo]
+----
+function counter = |init, stepFun| {
+  let current = box(init)
+  return -> current: getAndUpdate(stepFun)
+}
+#...
+let c2 = counter(3, |x| -> x * 2)
+c() # -> 3
+c() # -> 6
+c() # -> 12
+----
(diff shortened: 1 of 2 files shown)


danielpetisme:
`ref` is it the appropriate name ? Reading the PR title I thought it was something link to `fun` or `^`.
Wondering if couldn't be more explicit ? Simply `atomicRef` ?

jponge:
How about `box`? Otherwise `atomicRef` is better than `ref` indeed.

Another thing: you should add a quick documentation snippet in `doc/predefined-functions.asciidoc`.

yloiseau:
The name was inspired by OCaml. I wanted something short and explicit. From my point of view, the atomic nature is not what matters, it's just a mutable reference to an object (although the atomicity is important). I prefer `box` over `atomicRef`.
Should it be `Box` or `box` ? It's a function, but used as a constructor `let a = Box(42)` vs. `let a = box(42)` (btw, we could say the same about `range` vs. `Range` :smile:)

OT, I just noted that `reversed_range` is the only public function not named camelCase... Don't know why I used underscore notation, but maybe we should fix that (`reversedRange`, keeping a deprecated `reversed_range`), for consistency, or we will end up looking like PHP :smile:

jponge:
You'd have my vote for `Box` :smiley: 

Also :+1:  on `reversedRange` over `reversed_range`

yloiseau:
Ok, so `Box` (upper), `range` (lower), and `reversedRange` (mixed)? This lead to inconsistency between `Box` and `range`, which are both “constructor-like” functions named alike the corresponding “type” (as opposed to factories that have more explicit names, like `reversedRange` or java `valueOf`).

jponge:
`box` then

yloiseau:
Ok. Maybe I over-think about all this :smile:

yloiseau:
on the other hand, we have `Predefined.mapEntry`, so lower-case is predominant…

jponge:
I really believe that `box` is just fine and readable.

doc/predefined-functions.asciidoc
```diff
@@ -297,3 +297,20 @@ let e = mapEntry("foo", "bar")
 # prints "foo => bar"
 println(e: getKey() + " => " + e: getValue())
 ----
+
+`box` gives instances of `java.util.concurrent.atomic.AtomicReference`, and can
+be used to create a muttable reference where not possible othewise, e.g. in
```
jponge: `mutable`, not `muttable` :wink:

jponge:
@yloiseau I'll let you merge after editing the typo.

PR #302, opened 2015-08-25
Fix: inconsistent naming for reversed_range

yloiseau:
rename `reversed_range` to `reversedRange`, and create a new
*deprecated* `reversed_range` delegating on it.


jponge:
All good :+1:

PR #306, opened 2015-10-08
Code quality: fix findbugs and checkstyle.

yloiseau:
Fixes most of the issues raised by findbug.
Also clean some code according to checkstyle.

Need some review, since I'm not 100% sure that the fix was the way to go.


gradle/findbugs-exclusions.xml
```diff
@@ -10,4 +10,13 @@
   <Match>
     <Class name="~org\.eclipse\.golo\.compiler\.parser.*" />
   </Match>
+  <Match>
+    <!-- Ignore naming conventions for methods and classes used in golo code -->
+    <Class name="~gololang.*" />
+    <Bug code="Nm" />
+  </Match>
+  <Match>
+    <Class name="gololang.Unknown" />
```
jponge: Why `Unknown`? What can it match?

src/main/java/gololang/FunctionReference.java
```diff
@@ -49,7 +49,11 @@ public FunctionReference(MethodHandle handle, String[] parameterNames) {
       throw new IllegalArgumentException("A method handle cannot be null");
     }
     this.handle = handle;
-    this.parameterNames = parameterNames;
+    if (parameterNames != null) {
```
jponge: I'm unsure a defensive copy here is helpful. In regular Java code I would say yes, but here we have an array that is kept private, and the values come from the runtime bootstrap.

The case of `null` is good, but again, is it worth allocating an array?

src/main/java/gololang/Predefined.java
```diff
@@ -486,13 +486,32 @@ private static Path pathFrom(Object file) {
...
+
+  /**
+   * Writes some text to a file using the given {@link Charset}.
+   * The file is created if it does not exist, and overwritten if it already exists.
+   *
+   * @param text the text to write.
+   * @param file the file to write to as an instance of either {@link String}, {@link File} or {@link Path}.
+   * @param charset the charset to encode the text in.
+   */
+  public static void textToFile(Object text, Object file, Object charset) throws Throwable {
```
jponge: Good that it keeps the existing APIs and 2-arity `textToFile` above.

src/main/java/gololang/Tuple.java
```diff
@@ -99,6 +100,9 @@ public boolean hasNext() {
 
       @Override
       public Object next() {
+        if (i >= data.length) {
```
jponge: :+1:

gradle/findbugs-exclusions.xml
jponge: Ah ok, I had forgotten about that class...

src/main/java/org/eclipse/golo/cli/command/InitCommand.java
```diff
@@ -111,14 +112,18 @@ private String escapeModuleName(String projectName) {
...
   private void writeProjectFile(File intoDir, String projectName, String sourcePath, String fileName) throws IOException {
     InputStream sourceInputStream = InitCommand.class.getClassLoader().getResourceAsStream(sourcePath);
-    BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(sourceInputStream));
+    BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(sourceInputStream, StandardCharsets.UTF_8));
     File projectFile = new File(intoDir, fileName);
     PrintWriter writer = new PrintWriter(projectFile, "UTF-8");
     String line;
-    while ((line = bufferedReader.readLine()) != null) {
-      writer.println(line.replace("{{projectName}}", projectName));
+    try {
```
jponge: :+1:

src/main/java/org/eclipse/golo/compiler/CodeGenerationResult.java
```diff
@@ -27,7 +29,11 @@
    * @param packageAndClass the package and class descriptor for the bytecode.
    */
   public CodeGenerationResult(byte[] bytecode, PackageAndClass packageAndClass) {
-    this.bytecode = bytecode;
+    if (bytecode == null) {
```
jponge: Again, this is all compiler-internal, and I'm not sure it helps. But since it is not performance-criticial compared to the runtime I would say yes for this one :smile:

src/main/java/org/eclipse/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -183,7 +183,7 @@ private void writeAugmentationApplicationsMetaData(Map<String, Set<String>> appl
     int[] keys = new int[applicationsSize];
     String[][] namesArrays = new String[applicationsSize][];
     // cases of the switch statement MUST be sorted
-    Collections.sort(applicationNames, new Comparator<String>(){
+    Collections.sort(applicationNames, new Comparator<String>() {
```
jponge: A lambda would be possible here for readability.

src/main/java/org/eclipse/golo/compiler/JavaBytecodeStructGenerator.java
```diff
@@ -55,7 +55,8 @@ private void makeSetMethod(ClassWriter classWriter, Struct struct) {
     for (String member : struct.getMembers()) {
       visitor.visitLdcInsn(member);
       visitor.visitVarInsn(ALOAD, 1);
-      visitor.visitJumpInsn(IF_ACMPNE, nextCase);
+      visitor.visitMethodInsn(INVOKEVIRTUAL, "java/lang/String", "equals", "(Ljava/lang/Object;)Z", false);
```
jponge: Oh... :+1:

gradle/findbugs-exclusions.xml
yloiseau: It's a special singleton added for special matching of union types, to represent values you don't care about (like _ in Caml and other languages). e.g.
```golo
let _ = Unknown: get()
union Tree = {
  Node = {left, right}
  Leaf = {value}
  Empty
}

match {
  when tree: isEmpty() then "empty tree"
  when tree: isLeaf(0) then "a leaf with 0"
  when tree: isLeaf() then "a leaf"
  when tree: isNode(Tree.Empty(), _) or tree: isNode(_, Tree.Empty()) then "node with 1 child"
  when tree: isNode() then "a node"
  otherwise "wtf"
}
```

To work, the instance of `Unknown` is equals to anything but null. But findbugs does not like that, so I mask it out.

src/main/java/gololang/FunctionReference.java
yloiseau: If I recall correctly, findbugs was coplaining that we expose the internal of the object by taking or returning a mutable structure, breaking encapsulation. So I copied it :)
Maybe we can in that case simply mute the error

src/main/java/org/eclipse/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
yloiseau: indeed...

src/main/java/gololang/FunctionReference.java
yloiseau: Should I revert and mute the findbug error?

jponge: This one yes :smiley:

src/main/java/org/eclipse/golo/cli/command/InitCommand.java
```diff
@@ -111,14 +112,18 @@ private String escapeModuleName(String projectName) {
...
+      while ((line = bufferedReader.readLine()) != null) {
+        writer.println(line.replace("{{projectName}}", projectName));
+      }
+    } finally {
```
Artpej: No sure, but isn't it possible to use a "try with resources" here?

jponge: Oh yes, it should be possible! Good catch.

yloiseau:
There are still 6 issues in findbugs:
- 5 about using `doPrivileged` blocks with class loaders (dunno what to do, I'm not that much into class loader magic incantations);
- 1 about using `System.exit()` while handling compilation exceptions in the CLI code:
   findbugs suggest to throw a `RuntimeException` instead… I rather suggest to silent this error.

jponge:
Regarding the `doPrivileged` blocks I'd rather avoid dealing with the `SecurityManager`. I'm pretty sure we break some rules.

I also agree on silencing the `System.exit()` cases.

yloiseau:
Ok to merge this one ?

jponge:
Yes, thanks Yannick!
Le mer. 14 oct. 2015 à 16:28, Yannick Loiseau <notifications@github.com> a
écrit :

PR #307, opened 2015-10-09
Make the range classes use a static inner class

jponge:
This is generally better to use static vs instance-attached inner classes (performance, GC, etc).

Another point is that this makes the Eclipse compiler happy due to a bug it has.

### DIFF
--- a/src/main/java/gololang/CharRange.java
+++ b/src/main/java/gololang/CharRange.java
@@ -37,10 +37,10 @@ final class CharRange extends AbstractRange<Character> {
 
   @Override
   public int size() {
-    if (to() == from()) {
+    if (to().equals(from())) {
       return 0;
     }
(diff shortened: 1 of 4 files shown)


jponge:
Ping @yloiseau for review + merge

yloiseau:
My bad...

src/main/java/gololang/CharRange.java
```diff
@@ -37,10 +37,10 @@ Character defaultValue() {
 
   @Override
   public int size() {
```
yloiseau: Couldn't we now generalize by moving `size` into `AbstractRange`

jponge: Indeed!

Le dim. 11 oct. 2015 à 11:23, Yannick Loiseau <notifications@github.com> a
écrit :

PR #309, opened 2015-10-27
Refactor IR and its construction

yloiseau:
This is a refactoring to ease macro implementations.

- makes IR nodes mutable (in a limited way)
- makes a builder-like fluent interface to IR nodes
- better encapsulation by moving methods in the appropriate node
- adds a walk method to node to ease the visit (internal iteration)
- abstract dummy IR visitor to ease the creation of new specific
  visitors
- refactoring of IR visitors to use the abstract one
- introduction of new IR nodes (e.g. augmentations,
  ForEachLoop) to move some syntactic sugar from AST to IR transformation
  into a separate IR visitor (this will allow macros to build such node
  more easily)
- adds Builders to create a DSL to ease the creation of the IR
- refactoring of AST to IR visitor to use the new DSL and fluent API

Also clean some code according to PMD and checkstyle


jponge:
This is big one Yannick :-)

I’ll have a look soon.

Thanks!

jponge:
I'm doing a quick pass now.

src/main/java/gololang/Observable.java
```diff
@@ -109,15 +109,15 @@ public String toString() {
         '}';
   }
 
-  public static interface Function {
+  public interface Function {
```
jponge: Why did you remove `static` from the interface definitions?

src/main/java/gololang/Predefined.java
```diff
@@ -598,7 +603,7 @@ public static boolean isArray(Object object) {
    * @return the Character value.
    * @throws IllegalArgumentException if {@code obj} is not a number or a String.
    */
-  public static Object charValue(Object obj) throws IllegalArgumentException {
+  public static Object charValue(Object obj) {
```
jponge: Why removing the `throws` declarations?

src/main/java/org/eclipse/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -79,56 +52,89 @@
...
+    throw invalidElement(match);
+  }
+
+  @Override
+  public void visitCaseStatement(CaseStatement caseStatement) {
+    throw invalidElement(caseStatement);
+  }
+
+  @Override
+  public void visitWhenClause(WhenClause<?> whenClause) {
```
jponge: Why is there a generic?

src/main/java/org/eclipse/golo/compiler/JavaBytecodeGenerationGoloIrVisitor.java
```diff
@@ -220,34 +228,47 @@ private void writeAugmentationApplicationsMetaData(Map<String, Set<String>> appl
...
+        imports.stream()
+        .map(ModuleImport::getPackageAndClass)
+        .map(PackageAndClass::toString)
+        .toArray(String[]::new));
   }
 
-  private void writeAugmentsMetaData(Set<String> augmentations) {
-    String[] augmentArray = augmentations.toArray(new String[augmentations.size()]);
-    writeMetaData("augmentations", augmentArray);
+  private void writeAugmentsMetaData() {
```
jponge: :+1: :smiley:

src/main/java/org/eclipse/golo/compiler/ir/ConstantStatement.java
```diff
@@ -11,9 +11,9 @@
 
 public class ConstantStatement extends ExpressionStatement {
 
-  private final Object value;
+  private Object value;
 
-  public ConstantStatement(Object value) {
+  ConstantStatement(Object value) {
```
jponge: Why not `public` anymore?

src/main/java/org/eclipse/golo/compiler/ir/WhenClause.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.golo.compiler.ir;
+
+public final class WhenClause<T extends GoloElement> extends GoloElement {
```
jponge: Ok, got it wrt previous comment.

src/main/java/gololang/Predefined.java
yloiseau: Not mandatory since it's an unchecked exception (pmd was complaining). But it can indeed be useful as a documentation purpose

src/main/java/org/eclipse/golo/compiler/ir/ConstantStatement.java
yloiseau: To "force" the use of the corresponding `Builders` method

src/main/java/gololang/Observable.java
yloiseau: It's redundant (a nested interface is automatically "static")

src/main/java/gololang/Predefined.java
jponge: > Not mandatory since it's an unchecked exception (pmd was complaining). But it can indeed be useful as a documentation purpose

In that case PMD is wrong :smile:

(note that I don’t like PMD)

src/main/java/org/eclipse/golo/compiler/ir/ConstantStatement.java
jponge: Ok

src/main/java/gololang/Observable.java
jponge: Oh yes, right!

src/main/java/gololang/Predefined.java
yloiseau: Just curious, why do you say it's wrong? Because of the documenting value of being explicit?
If so I agree with that point, although is in the javadoc :smile:
Anyway, I'll put it back. I've removed such explicit `throws` of `RuntimeExceptions` in several places, should I rollback everywhere?

jponge: Yes, I value the explicitness for unchecked exceptions that may still likely be thrown. Javadoc is one thing, yet `throws` does, too.

The exception is probably `RuntimeException` which is very generic, but all of that is me :smile:

jponge:
@yloiseau Looks good for me

PR #310, opened 2015-10-29
Fix: the same parsing error appears multiple times

Artpej:
The same parsing error appears multiple times.

as example:

```golo
module test

function main = |args| {
  # missing double quote
  let a = "aBc
}
```

prints 3 times the same error:

```
[error] In Golo module: test.golo
[error] Encountered unexpected  <INVALID> `\"`  at line 5, column 11
[error] Encountered unexpected  <INVALID> `\"`  at line 5, column 11
[error] Encountered unexpected  <INVALID> `\"`  at line 5, column 11
```

### DIFF
--- a/src/main/jjtree/org/eclipse/golo/compiler/parser/Golo.jjt
+++ b/src/main/jjtree/org/eclipse/golo/compiler/parser/Golo.jjt
@@ -40,11 +40,16 @@ public class GoloParser {
 
   public GoloCompilationException.Builder exceptionBuilder;
 
+  private boolean errorAlreadyReported = false;
+
   void skipto(int kind, ParseException e, GoloASTNode node) throws ParseException {
     if (exceptionBuilder == null) {
       throw e;
     }
-    exceptionBuilder.report(e, node);
+    if (!errorAlreadyReported) {
+      exceptionBuilder.report(e, node);
+      errorAlreadyReported = true;
+    }
     if ("<EOF>".equals(e.tokenImage[0])) {
       return;
     }


Artpej:
There's maybe a better way to fix this but this one seems do the job...
Any clue everybody?

jponge:
What if you have another parsing error?

jponge:
I haven't tested, but I have a suspicion that it masks other errors :smile:

Artpej:
Only parsing errors (javacc) are problematic.
I thought that when javacc find an error, then it stops immediatelly without report subsequent errors.
Do you have an example where javacc reports more than one error?

Antoher way to solve this could be to check that the `Problem` is not already reported in the problem stack (overrirding `hashCode` and `equals` for exemple) but I think it's a little bit <del>ugly</del> tricky :smile:

thoughts?

jponge:
You may have 2 errors to report, for instance you have `def main` instead of `function main`.

Artpej:
I'm not sure i undertand :wink: 

this one:
```golo
module test

def main = |args| {
  # missing double quote
  let a = "aBc
}
```
just prints this and stop:
```
[error] In Golo module: test.golo
[error] Encountered unexpected  <IDENTIFIER> `def`  at line 3, column 1
``` 

and this one
```golo
module test

let main = |args| {
  # missing double quote
  let a = "aBc
}
```
prints this and stop:
```
[error] In Golo module: test.golo
[error] Encountered unexpected  <INVALID> `\"`  at line 5, column 11
[error] Encountered unexpected  <INVALID> `\"`  at line 5, column 11
[error] Encountered unexpected  <INVALID> `\"`  at line 5, column 11
```

jponge:
I guess it depends on the parser recovery :-)

jponge:
Hum did some checks, and you are right, parser errors are always single it seems.

jponge:
So yes @Artpej you have a point. That being said the `boolean` flag seems a bit dirty to me... :smile:

jponge:
I think you get 3 errors due to the newline, the `}` and `<EOF>`

jponge:
Merging this, thanks @Artpej.

We'll possibly rework the error handling in the future.

PR #311, opened 2015-11-11
Change implicit import order

yloiseau:
Import explicit modules *before* implicit ones (e.g.
`gololang.StandardAugmentations`) such that we can for instance redefine
standard augmentations.

When a type is augmented several times with the same method, the first one is used, that is the one defined in the first imported module.
This is ok, since you can “override” an augmentation by changing the import order. However, since some modules are implicitly imported first (e.g. `StandardAugmentations`), it is not possible to redefine augmentations specified in these modules, i.e. one can’t redefine standard augmentations. 

This limit the usefulness of augmentations. For instance, suppose one want to define a `reduce` method on a specific `struct`. Since a `struct` is a subclass of `GoloStruct`, that implements `Iterable`, and since `Iterable` already has a `reduce` augmentation in `StandardAugmentations`, one can’t define this specific method on the `struc`. This solution simply import implicit modules after the explicit ones. 
Thus, the “user defined” augmentations defined in explicitly imported modules are used, the standard augmentations beeing used as a “fallback” if no other augmentation is defined in the current (or imported) module.

### DIFF
--- a/doc/augmentations.asciidoc
+++ b/doc/augmentations.asciidoc
@@ -251,7 +251,7 @@ The augmentations resolution order is as follows:
    using the same order than locally applied ones, in the importation order.
 
 The first matching method found is used. It is thus possible to “override” an
-augmentation with a more higher priority one (in the sens of the previous order). xref:implicit_imports[Implicit modules] are imported after explicit ones to allow to redefined standard augmentations.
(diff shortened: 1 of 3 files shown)


jponge:
Sorry for the title changes, my son was playing around :smile:

yloiseau:
:smile:

jponge:
Looks good. We should document this behaviour, what do you think?

yloiseau:
Indeed.

jponge:
:+1: thanks!

PR #312, opened 2015-11-13
New syntax to specify arity in function literals

yloiseau:
Introduces a extended notation for function reference literals that
allows to specify the arity of the function.

The general syntax is `^package.module.::function\arity`

Example:

```golo
function foo = |a, b| -> a * b

function foo = |x| -> x + 1

function call_foo = {
  return ^foo\2(^foo\1(20), 2)
}
```

### DIFF
--- a/doc/closures.asciidoc
+++ b/doc/closures.asciidoc
@@ -283,6 +283,17 @@ function call_foo = {
 }
 ----
 
+In this case, one can even specify if a varargs function is needed or not (since arity alone can be ambiguous) using final `...`:
+
+[source,golo]
+----
+local function foo = |a| -> "unary"
+local function foo = |a...| -> "variable"
+
+^foo\1(null) # unary
+^foo\1...()  # variable
+----
+
 Note that you can in the same way get a reference to a Java static method. For
 instance:
 
--- a/doc/closures.asciidoc
+++ b/doc/closures.asciidoc
@@ -269,6 +269,20 @@ function call_local_fun = {
 }
 ----
 
+If the function is overridden, that is several functions exist with the same name but different arities, 
+you *must* specify the arity with the notation `^moduleName::functionName\arity`. For instance:
+
+[source,golo]
+----
+local function foo = |a, b| -> a * b
+
+local function foo = |x| -> x + 1
+
+function call_foo = {
+  return ^foo\2(^foo\1(20), 2)
+}
+----
+
 Note that you can in the same way get a reference to a Java static method. For
 instance:
 
--- a/src/main/java/org/eclipse/golo/compiler/SugarExpansionVisitor.java
+++ b/src/main/java/org/eclipse/golo/compiler/SugarExpansionVisitor.java
@@ -165,7 +165,7 @@ class SugarExpansionVisitor extends AbstractGoloIrVisitor {
             classRef(ref.module == null
               ? this.module.getPackageAndClass()
               : ref.module),
-            constant(-1));
+            constant(ref.arity));
       constantStatement.replaceInParentBy(fun);
       fun.accept(this);
       return;
--- a/src/main/java/org/eclipse/golo/compiler/SugarExpansionVisitor.java
+++ b/src/main/java/org/eclipse/golo/compiler/SugarExpansionVisitor.java
@@ -165,7 +165,8 @@ class SugarExpansionVisitor extends AbstractGoloIrVisitor {
             classRef(ref.module == null
               ? this.module.getPackageAndClass()
               : ref.module),
-            constant(ref.arity));
+            constant(ref.arity),
+            constant(ref.varargs));
       constantStatement.replaceInParentBy(fun);
       fun.accept(this);
       return;
--- a/src/main/java/org/eclipse/golo/compiler/ir/Builders.java
+++ b/src/main/java/org/eclipse/golo/compiler/ir/Builders.java
@@ -180,7 +180,15 @@ public final class Builders {
   }
 
   public static ConstantStatement functionRef(Object moduleName, Object funcName, Object arity) {
-    return constant(new GoloParser.FunctionRef((String) moduleName, (String) funcName, (Integer) arity));
+    return functionRef(moduleName, funcName, -1, false);
+  }
+
+  public static ConstantStatement functionRef(Object moduleName, Object funcName, Object arity, Object varargs) {
+    return constant(new GoloParser.FunctionRef(
+          (String) moduleName,
+          (String) funcName,
+          (Integer) arity,
+          (Boolean) varargs));
   }
 
   public static ReturnStatement returns(Object expr) {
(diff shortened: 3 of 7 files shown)


yloiseau:
Just a little issue with this... It's just minimal syntactic sugar, which generates a call to `Predefined.fun`, and I didn't want to change the `fun` behavior, but this lead to some strange cases, when dealing with non-static Java methods. The current behavior is to return the unbound method, that is a function taking the object itself as first parameter. However, the declared arity is 0, which leads for instance to
```golo
^String::toUpperCase\0("hello")
```
(since `toUpperCase can have a locale as parameter). This can be misleading. Note that this is the same with `fun`:
```golo
fun("toUpperCase", String.class, 0)("hello")
```

My fault, I introduced the change to allow non-static methods, but didn't envision the arity issue.
BTW, what should we do?

* specify the expected arity and add a special case in `fun` to decrement it if the method is not static;
* leave it as is, and document this strange behavior.

danielpetisme:
the synax `\{arity}` is a bit tricky, IMHO, I fell like reading low-level code in a high-level language. I do prefer the verbose one.... And why not simply add a flag to the `fun` method
```
funtion fun = |name, type, arity, isInstance|
function fun = |name, type, arity| -> fun(name, type, arity, false)
```

yloiseau:
The syntax is inspired what can be found in languages like prolog, or more recently elixir (which are very high level languages :wink:)

yloiseau:
An other solution would be to redefine `fun` not to return a unique function reference of fail if multiple solution (when no arity is specified), but to return a “proxy” function reference, containing all the possible alternatives, and resolving the correct one when called. For instance, given
```golo
function foo = |x| -> ...
function foo = |x, y| -> ...

let f = ^foo
# or
let g = fun("foo", currentModule.module)
```
`f` (or `g`) would be a proxy containing both function references, and calling `f(1)` would dispatch to the first one, while `f(1, 2)` would dispatch to the second one. However, I think that “explicit is better than implicit”, and besides, it could be really tricky to do if one of the function has variable arity
```golo
function foo = |x...| -> ...
function foo = |x, y| -> ...
let f = ^foo
```
we can't determine which method `^f(1, 2)` must call.

yloiseau:
By the way… while I was thinking about that, I just realized that even the behavior of normal calls is not really defined in this case… and BLAM!
```golo
module Test

function foo = |x...| -> "v"

function foo = |x, y| -> "d"

function main = |args| {
  if foo(1, 2) == "d" {
    println("ok")
  } else {
    println("FAILED")
  }
}
```

```bash
$ while true; do golo golo --files test.golo; done
ok
ok
ok
FAILED
ok
ok
ok
ok
FAILED
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
ok
FAILED
ok
FAILED
ok
FAILED
FAILED
ok
FAILED
ok
ok
ok
ok
ok
ok
ok
ok
FAILED
FAILED
ok
ok
ok
FAILED
```

some stats (only printing when failed)
```bash
$ for i in `seq 100`; do golo golo --files test.golo ; done | wc -l
30
```

while doing the loop in the golo code
```golo
function main = |args| {
  foreach i in [0..100] {
    if foo(i, i+1) == "v" {
      println("FAILED")
    }
  }
}
```
due to inline cache, it always OK, or never:
```bash
$ while true; do golo golo --files test.golo | wc -l ; done
0
100
100
0
100
0
100
100
0
0
0
100
100
100
100
100
0
0
100
0
100
^C
```

ouch… new bug to fill

Artpej:
I agree with @danielpetisme, this improvment is useful but it seems to me adding somthing new in the grammar is not needed.

I throught about somthing more golo freindly in order to deal with different arity functions. 
`^foo` returns a `FunctionReference` for `foo`. If there's multiple `foo` functions we could return something like `MultipleFunctionReference` and resolve the more appropriate one at call time (indy powa :wink: ).

`MultipleFunctionReference` could have sevral methods for deal with different arity and so on...

example:
```golo
function foo = |a, b| -> a * b

function foo = |x| -> x + 1

function main = |args| {
  #get the MultipleFunctionReference
  let f = ^foo
 
 # dispatch to the first one
  f(1,2)

  # dispatch to the second one
  f(1)

  # and with methods
  if f: isMultilpe() {
    # call the first one
    f:fromArity(2)(1,2)
    # call the second one
    f:fromArity(1)(1)
  }
}
```
 My two cents, WDYT?

yloiseau:
I evoked this solution in [my previous comment](#issuecomment-157096310) :smile:
But how do you dispatch if you have 
```golo
function foo = |x...| -> ...
function foo = |x, y| -> ...

^foo(1, 2)
```

Artpej:
@yloiseau  arggg... my eyes are not totally opned before my first coffe :smile:

we can warn (or throw an error) as compile time that the 2 first funtions can ben hide by the third one as mentioned in the #313 . In the case of just a warn we could define a priority (for example fixed arity is always priorised on variable arity), in this case i don't know if the warn is still needed.

yloiseau:
As for the PR, I rather write
```golo
coll: map(^foo\1)
```
than
```golo
coll: map(fun("foo", MyLocalModule.module, 1))
```

yloiseau:
I also “discovered” that `fun` can access *local functions* in a module. Bug or feature ?

jponge:
1. On `fun` accessing local functions: *bug* if this is possible from _outside_ the module, *feature* if it just works from _inside_ the module.
2. I don't mind the `^plop\6` syntax.
3. A `fun` variant could indeed have an arity parameter.

yloiseau:
@jponge so we have a bug :smile:

```golo
module Other

local function bar = |x| -> x
```
```golo
module Test

local function foo = |x| -> x

function main = |args| {
  let f = fun("foo", Test.module)
  println(f(42))

  let b = fun("bar", Other.module)
  println(b(42))
}
```

```bash
$ golo golo --files other.golo test.golo 
42
42
```

jponge:
Nice spot!

jponge:
Just created https://bugs.eclipse.org/bugs/show_bug.cgi?id=482455

yloiseau:
Let's continue the discussion in ambiguous function references on the golo-dev ML since the scope is larger than this PR...

jponge:
Any news?

yloiseau:
Since we have a problem with varargs, I'm waiting for @Artpej's work on https://github.com/eclipse/golo-lang/pull/313#issuecomment-162469641

yloiseau:
waiting for #326

yloiseau:
Since #366 fixed some issues, I think I'm OK with this one

jponge:
@yloiseau Do you mean it's good to merge now?

yloiseau:
yep, if everyone is OK

yloiseau:
oops, need to update the doc...

jponge:
Good to merge!

PR #313, opened 2015-11-15
Allow to create custom factories for structs

Artpej:
With this, the Factories generated by Golo can be overloaded by custom ones.

for example:

```golo
module test

struct Point = { x, y }

function Point = -> test.types.Point(0,0)

function main = |args| {
  println(Point(1, 2))
  println(Point())
}

```
running it prints the following console output:

```
struct Point{x=1, y=2}
struct Point{x=0, y=0}
```


src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
```diff
@@ -334,4 +335,30 @@ public void walk(GoloIrVisitor visitor) {
...
   }
+
+  @Override
+  public int hashCode() {
+    int hash = 5;
+    hash = 89 * hash + Objects.hashCode(this.name);
+    hash = 89 * hash + Objects.hashCode(this.getArity());
+    return hash;
+  }
+
```
yloiseau: why don't you just use `Objects.hash(name, getArity())`?

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
```diff
@@ -334,4 +335,30 @@ public void walk(GoloIrVisitor visitor) {
...
+    }
+    final GoloFunction other = (GoloFunction) obj;
+    if (!Objects.equals(this.name, other.name)) {
+      return false;
+    }
+    if (this.getArity() != other.getArity()) {
+      return false;
+    }
+    return true;
+  }
```
yloiseau: `return Objects.equals(this.name, other.name) && this.getArity() == other.getArity()` ? :smile:

src/main/java/org/eclipse/golo/compiler/SugarExpansionVisitor.java
```diff
@@ -315,4 +315,14 @@ public void visitDestructuringAssignment(DestructuringAssignment assignment) {
...
+
+  /**
+   * Add struct factories if they don't already exists
+   *
+   */
+  @Override
+  public void visitStruct(Struct struct) {
+    module.addFunctions(struct.createFactories());
+  }
+
```
yloiseau: My previous refactoring helped, didn't it? :wink:

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
Artpej: We have to check that `other` is not `null` and that it is a `GoloFunction` instance :wink:

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
Artpej: No idea, should I be less confident with my IDE that do `hashCode` automatically like this?

src/main/java/org/eclipse/golo/compiler/SugarExpansionVisitor.java
Artpej: indeed! :wink:

yloiseau:
I think more criteria should be used to `equals`/`hashCode` functions, e.g. at least scope, maybe visibility.

For instance:
```golo
module Plop 

struct Foo = { x, y }
augment Foo {
  function plop = |this| -> "ploping: " + this: x() + ", " + this: y()
}

function plop = |val| -> "Plop with " + val
```
the two `plop` functions would be considered equals. Not an issue for your change, but I think it can latter leads to strange behavior.

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
yloiseau: Yes but I mean after the cast, you can just return the boolean expression instead of the two `if`s (since `&&` is lazy)

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
yloiseau: I think it's functionally equivalent, it just seems more readable to me.

danielpetisme:
I <3 this PR, 
is it out of the box compatible with Immutable structs ?

yloiseau:
It works by *not* injecting the factory methods in the module if an equal one (same name and arity) is present, so yes, if you define an `ImmutableMyStruct` with the right arity, it will “override” the default one.

yloiseau:
This will indeed allow to check preconditions on instanciation :smile:.
Again about not taking account of visibility in the `equals`, it can be a feature (or not). Indeed, if you do:
```golo
module Test

struct MyStruct = {x, y}

local function MyStruct = |x, y| -> Test.types.MyStruct(x, y)
local function MyStruct -> Test.types.MyStruct(0, 0)
```
the local function will be considered equal to default factories, and thus they will not be injected. That is, you can't create a `MyStruct` like:
```golo
module Use

import Test

function main = |args| {
  let s = MyStruct(1, 2) # will fail
  let t = Test.types.MyStruct(1, 2) # should be ok
}
```
It can be seen as a feature :wink:

Artpej:
@danielpetisme there's 2 ways to "override" the immutables factories:

#### With a frozen copy
```golo
module test

struct Point = { x, y }

function Point = -> test.types.Point(0,0)

function ImmutablePoint = |a,b| -> test.types.Point(a,b): frozenCopy()

function main = |args| {
  var point = Point(1, 2)
  println(point + " " + point: isFrozen())
  point = ImmutablePoint(1,2)
  println(point + " " + point: isFrozen())
}
```

#### Or with internals knowleges :wink:
```golo
module test

struct Point = { x, y }

function Point = -> test.types.Point(0,0)

function ImmutablePoint = |a,b| -> test.types.Point.$_immutable(a,b)

function main = |args| {
  var point = Point(1, 2)
  println(point + " " + point: isFrozen())
  point = ImmutablePoint(1,2)
  println(point + " " + point: isFrozen())
}

```

@yloiseau i think function name + arity is suffisent for function discrimination inside a module. The unwanted behaviour is that if we declare tow times a function with same name and arity only the first declared one will be used without warn at compilation time. (before this PR an exception was thrown at runtime: `Exception in thread "main" java.lang.ClassFormatError: Duplicate method name&signature in class file`)

Artpej:
@yloiseau ok I just had a look on #312 comments and we should take in account variable arity too :wink:

Maybe we could warn at compilation when a function is suceptible to hide another one (except for the struct factories since it's a feature :smile: )

yloiseau:
I think that at least a warning is needed. I'd argue it should be a *compile time* error. Checking can be as simple as:
```java
public void addFunction(GoloFunction function) {
    if (functions.contains(function) && !function.isSynthetic()){
        // throw ....
    }
    ...
  }
```

since generated default factories are synthetic, they are just ignored, but a user defined function would failed.

yloiseau:
see https://gist.github.com/yloiseau/464e283d85e2ae3d91dd, I created that to test some thing while refactoring IR (don't remember what), but did'nt used it. 
Maybe instead of overriding `GoloFunction.equals`, the module functions can be a `Map<FunctionSignature, GoloFunction>`

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
jponge: IDE do it right usually, especially when you specify which fields are nullable :+1:

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
jponge: Same here, IDEs do it right.

But Java 8 `Object.equals` is also fine.

jponge:
Very good pull-request!

jponge:
I concur with @yloiseau that function hiding shall raise a compilation error.

jponge:
Assigned to @Artpej, please ping us when a new review round is possible :smile:

yloiseau:
Let's continue the discussion on ambiguous function references on the golo-dev ML since the scope is larger than this PR...

jponge:
Ping, any news?

Artpej:
Still working one this :
 - Add compilation error when a function hide another one (fixed arity)
 - keep the same function resoluion order as Java for varargs ([reference](http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2.4))

I will try to push an update for this proposal by the end of the week.

jponge:
:+1:

Artpej:
I just updated this one: 

- [x] Add compilation error when a function hide another one (fixed arity)
- [x] fixed arity function are now resolved before varargs one 

I think about a huge refactoring for functions resolution based on works that are logged here  #326  thanks to @yloiseau :wink: but it's more complexe work than this one...

jponge:
Thanks @Artpej 

Are we good to go on this one?

Artpej:
Seem's to me fully fonctionnal like this. It is not as elegant as I expected, but a refactoring is coming. 
Ready for review.

danielpetisme:
I'm really impatient to use this feature for the testing module, it will really help!
I was bored to have a "init" method to override the no-args constructor. 
Thanks @Artpej

doc/structs.asciidoc
```diff
@@ -102,6 +102,57 @@ function main = |args| {
 }
 ----
 
+=== Custom factories
+
+Golo generates two factories for the structures. One with member's initial values as parameter and
```
jponge: for structures

doc/structs.asciidoc
```diff
@@ -102,6 +102,57 @@ function main = |args| {
...
+
+running it prints the following console output:
+
+[source,golo]
+----
+struct Point{x=1, y=2}
+struct Point{x=null, y=null}
+----
+
+NOTE: By default the no argument factory sets every member to `null`
```
jponge: no-argument

src/main/java/org/eclipse/golo/compiler/ParseTreeToGoloIrVisitor.java
```diff
@@ -120,7 +89,21 @@ public void leaveNamedAugmentation() {
     }
 
     public void addFunction(GoloFunction function) {
-      this.functionContainersStack.peek().addFunction(function);
+      FunctionContainer container = this.functionContainersStack.peek();
```
jponge: :+1:

jponge:
Done my review with little doc edit suggestions

Artpej:
done! :beers:

doc/structs.asciidoc
```diff
@@ -102,6 +102,57 @@ function main = |args| {
...
+
+The Factories generated by Golo can be overloaded by custom ones:
+
+[source,golo]
+----
+module test
+
+struct Point = { x, y }
+
+function Point = -> test.types.Point(0,0)
```
yloiseau: Maybe you could add a note on how to call the original constructor (the `<module>.types.<struct>` thing) and how to override the immutable constructor (see @danielpetisme question in the PR comments).

src/main/java/org/eclipse/golo/compiler/ParseTreeToGoloIrVisitor.java
```diff
@@ -120,7 +89,21 @@ public void leaveNamedAugmentation() {
...
+        GoloFunction firstDeclaration = null;
+        for (GoloFunction f : container.getFunctions()) {
+          if (function.equals(f)) {
+            firstDeclaration = f;
+          }
+        }
+        errorMessage(AMBIGUOUS_DECLARATION, function.getASTNode(),
+            String.format("Declaring a function `%s` twice (declared first at line=%d, column=%d) ",
+                function.getName(),
+                firstDeclaration.getASTNode().getLineInSourceCode(),
```
yloiseau: You can directly use `GoloASTNode::getPositionInSourceCode` which override `toString`...

src/main/java/org/eclipse/golo/compiler/ParseTreeToGoloIrVisitor.java
```diff
@@ -162,10 +145,36 @@ private LocalReference getOrCreateReference(ASTLetOrVar.Type type, String name,
...
+      }
+      return exceptionBuilder;
+    }
+
+    public void errorMessage(GoloCompilationException.Problem.Type type,
+                              GoloASTNode node,
+                              String message) {
+      String errorMessage = String.format(
+          "%s at (line=%d, column=%d)",
+          message, node.getLineInSourceCode(), node.getColumnInSourceCode());
```
yloiseau: same here, `GoloASTNode::getPositionInSourceCode`

src/main/java/org/eclipse/golo/compiler/ParseTreeToGoloIrVisitor.java
```diff
@@ -69,6 +37,7 @@ public Object visit(ASTerror node, Object data) {
     private Deque<FunctionContainer> functionContainersStack = new LinkedList<>();
     private Deque<Object> objectStack = new LinkedList<>();
     private Deque<ReferenceTable> referenceTableStack = new LinkedList<>();
+    private GoloCompilationException.Builder exceptionBuilder;
```
yloiseau: why not make it final and passed in the constructor (and remove the setter)?

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
```diff
@@ -344,4 +345,24 @@ public void walk(GoloIrVisitor visitor) {
...
     throw cantReplace();
   }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(this.name, this.getArity());
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
yloiseau: You don't test for `this == obj` ?

src/main/java/org/eclipse/golo/compiler/ir/GoloFunction.java
```diff
@@ -344,4 +345,24 @@ public void walk(GoloIrVisitor visitor) {
...
+      return false;
+    }
+    if (getClass() != obj.getClass()) {
+      return false;
+    }
+    final GoloFunction other = (GoloFunction) obj;
+    if (!Objects.equals(this.name, other.name)) {
+      return false;
+    }
+    return this.getArity() == other.getArity();
```
yloiseau: You don't test for varargs ? So
```golo
function foo = |a| -> ...
function foo = |a...| -> ...
```

will fail to compile.

src/test/resources/for-execution/structs.golo
```diff
@@ -80,3 +82,14 @@ function check_equals = ->
...
+struct AList = { _holder }
+
+augment golotest.execution.Structs.types.AList {
+  function add = |this, value| -> this: _holder(): add(value)
+  function size = |this| -> this: _holder(): size()
+}
+
+function AList = -> golotest.execution.Structs.types.AList(ArrayList())
+
+function check_overload_factory = -> AList(): size()
```
yloiseau: You don't have tests for the new behaviors you added (compilation error if existant function and varargs ordering in method resolution)

Artpej:
Thanks @yloiseau ! (my bad, functions are still resolved in `FunctionCallSupport` not in a `RegularMathodFinder`)

Artpej:
@yloiseau , oups I forgot change the `toString`. I'll ping you when it is done.

jponge:
@yloiseau will have the final word on the code review :smile: 

Great work Sylvain and Yannick!

yloiseau:
:+1:

Artpej:
@yloiseau ping!

yloiseau:
ok for me

jponge:
It's in, thanks @Artpej @yloiseau

jponge:
Ooops, it broke the build, any clue @Artpej?

Artpej:
I'll have a look asap

yloiseau:
@Artpej I posted a patch on gitter…

PR #315, opened 2015-11-20
Fix: issues when mixing module state and closures

yloiseau:
Their was some issues when mixing module state and closure:
- when a local closure used a module state value, it was captured, causing runtime errors
- when a module state containing a closure was defined after a regular functions, the reference tables was merged, causing compilation errors

This PR fixes both issues.


jponge:
:+1: :+1: :+1:

jponge:
Thanks Yannick

PR #325, opened 2015-12-11
Make structures comparable

yloiseau:
`GoloStruct` now implement `Comparable`, by comparing their members
pairwise in lexicographical order. `Tuple::compareTo` is used, doing
`this.values.compareTo(other.values())`

This allow to directly compare structures, using classical operators:

```golo
    struct Couple = { x, y }

    ...

    require(Couple(1, 2) < Couple(2, 4), "err")
```

An exception is raised if the structure are different, such that two
different structures with the same members are not comparable, for instance:

```golo
Couple(1, 2) < Point(2, 4)
```

will fail.


jponge:
Looks good to me!

PR #327, opened 2015-12-14
Operator call sites with catchException instead of a guardWithTest

jponge:
This proposed change invalidates operator call sites on `ClassCastException` rather than through a
branching `guardWithTest` method handle combinator. This removes 2 `instanceof` checks on each
call.

### DIFF
--- a/doc/structs.asciidoc
+++ b/doc/structs.asciidoc
@@ -199,7 +199,7 @@ will output:
 ----
 true
 true
-Exception in thread "main" java.lang.ClassCastException: struct Triplet{x=1, y=2, z=3} and struct Other{x=1, y=2, z=4} can't be compared
+Exception in thread "main" java.lang.IllegalArgumentException: struct Triplet{x=1, y=2, z=3} and struct Other{x=1, y=2, z=4} can't be compared
 ----
 
 even if `Triplet` and `Other` have the same members.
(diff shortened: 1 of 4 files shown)


jponge:
Ping @yloiseau: this changes your previous implementation of `struct` comparisons.

jponge:
@yloiseau Did you get a chance to review?

yloiseau:
Regarding the main idea, does not throwing exception is more expensive than checking the instance? I thought is was, but may be not, it's just a piori.

The implementation looks good to me.

On the change on struct, is does not strictly speaking comply to the `Comparable` interface anymore, since it should throw a `ClassCastException` (see [JDK](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html#compareTo-T-)), but since it's a runtime exception, I don't think it's much a big deal w.r.t. subtyping (in the sens of Liskov).

jponge:
There should be a fast path with no checks compared to a `if`, and it _should_ be faster. I suppose I should do a quick micro-benchmark to check :smile: 

I had missed the JDK contract that `ClassCastException` should be thrown. I'm willing to break this behaviour if the perf improves, and because I've never seen people have high expectations on this contract.

yloiseau:
Regarding the type of the exception, strictly speaking, it should be a subclass of `ClassCastException` to comply with output covariance. But this really matters when catching the exception, and since it's a `RuntimeException`, I think we can quite safely break it.

jponge:
Here is a quick microbenchmark and gross results: https://gist.github.com/jponge/82aa11193cc22f224ca7

For what it's worth, it seems like a good bet don't you think so?

yloiseau:
Worth breaking unchecked subtyping :smile:

jponge:
Merged!

Issue #235, opened 2015-01-13
Structure augmentations do not pass into the templates


k33g:
This is ok :

```golo
function ViewHelpers = {
  return DynamicObject()
    : define("plop", |this| {
        println("you're calling plop...")
        return "PLOP!!!"
    })
}

function helloViewTpl = -> """
<%@params entities, tools %>
<%= tools: plop() %>
<hr>
<h2><%= entities: get(0) %></h2>
"""

let helloView = gololang.TemplateEngine(): compile(helloViewTpl())
let tools = ViewHelpers()
helloView(list["Bob Morane"]  , tools)
```

But if I replace DynamicObject with augmented structure, `plop` method isn't seen when the compiled template is executed:

```golo
struct viewHelpers = {
  foo
}

augment viewHelpers {
  function plop = |this| {
    println("you're calling plop...")
    return "PLOP!!!"
  }
}

let helloView = gololang.TemplateEngine(): compile(helloViewTpl())
let tools = viewHelpers()
helloView(list["Bob Morane"]  , tools)
```

`Exception in thread "main" java.lang.NoSuchMethodError: class views.types.viewHelpers::plop`

yloiseau:
Is it a regression?

k33g:
In fact I don't know, it's the first time I try

k33g:
... the first try I test this use case

jponge:
It's because the augmentation is not visible from within the module.

Are you missing a `<%@import foo.bar.Baz %>` clause?

k33g:
oh oh, i'm going to try that after my meeting (my boss is sitting next to
me)

k33g:
I'm afraid that even with the import I've got a problem

jponge:
Then we need to investigate :-)

yloiseau:
It's also broken in v1.1.0

yloiseau:
BTW, it works on struct fields access. Only failed on augmentation...

k33g:
it's a feature ;) I'll use DynamicObjects instead

yloiseau:
?

yloiseau:
The template engine creates an anonymous `EvaluationEnvironment` to “privately” execute the generated code (sandbox like). Seems like the augmentations are not visible in this environment. 

Why do you say it’s a feature ?

jponge:
Augmentations shall be visible in that sandbox.

k33g:
@Yannick feature : golo humour ;)

[...]

Issue #326, opened 2015-12-12
Ambiguous functions declaration, overloading and shadowing


yloiseau:
Synthesis of comments on #312 #313 
Probably linked to #275 

We have an issue regarding the resolution of overloaded functions, more
specifically when using variable arity ones (varargs)

## A little history

When using the literal notation for function references
(`^module::function`) or the `Predefined::fun` function without
specifying the function arity (`fun("function", module.class)` or
`fun("function", module.class, -1)`), for which the literal notation is
just syntactic sugar, if the referenced function was overloaded, that is
the function has multiple definitions with different arities, the return
function reference used to be undefined. The first found was return, but
with no guaranty on with this first one would be. 

For instance, with code like 

    function foo = |x| -> ...

    function foo = |x, y| -> ...

    let f = fun("foo", currentModule.class)

`f` could be either of the two functions.

A correction was applied to raise
a `AmbiguousFunctionReferenceException` in this case, to avoid guessing.
This force the developer to be explicit on the required arity by using
`fun("foo", currentModule.class, 1)` or `fun("foo", currentModule.class,
2)`, instead of silently using one or the other.

The problem is not present when calling the function directly as in
`foo(1)` or `foo(1, 2)`, since in this case, the number of arguments is
known. This resolution takes place in
`runtime.FunctionCallSupport::findStaticMethodOrField`.

As a side note, their is some logic duplication between
`runtime.FunctionCallSupport::findStaticMethodOrField` and
`Predefined::fun` (and probably also in the method call resolution).

## Current problem

However, the direct call resolution use the number of *arguments* used
on the call site, and try to match them with the number of *parameters*
of the candidate function.

This raise an issue when the function is overloaded with a variable
arity that shadow the fixed one. For instance, given the functions:

    function foo = |x| -> ...

    function foo = |x, y| -> ...

    function foo = |xs...| -> ...

* when using `fun`, the third one shadows the first one, and thus
  `fun("foo", mod.class, 1)` throws an exception;
* when doing a direct call, the third one shadows the two others,
  depending on the number of arguments given: `foo(1)` can call the
  first or the third, and `fun(1, 2)` can call the second or the third.

## Some tests

See the two included files. This gave me:

    $ ./test.sh 1000
        379 direct 0: 0
        621 direct 0: 1v
        741 direct 1: 1
        151 direct 1: 1v
        108 direct 1: 2v
         93 direct 2: 1v
        799 direct 2: 2
        108 direct 2: 2v
[...]

PR #328, opened 2015-12-14
FIX #320: Lookup augmentations in call stack

### LINKED ISSUE #320: augmentations limited to imported scope
When an augmentation is defined, it is only visible if the module using it has imported the module defining it. This limits the ability to create generic functions.

For instance:
```golo
module MyData

struct MyStruct = { val }
augment MyStruct {
  function size = |this| -> this: val()
}
```

```golo
module MyLib

function printSize = |o| { 
  println("The size of " + o + " is " + o: size() + "!") 
}
```

```golo
module Main

import MyData
import MyLib

augment java.lang.String {
  function size = |this| -> this: length()
}

function main = |args| {

  # Ok, it's a native java method
  printSize(list[1, 2, 3])

  # Ok, DynamicObject
  printSize(DynamicObject(): define("size", |this| -> 42))

  # Fails since MyLib don't know the augmentation
  try {
    printSize("hello")
  } catch (e) {
    println("ERR: " + e)
  }

  # Fails since MyLib don't know the augmentation
  try {
    printSize(MyStruct(42))
  } catch (e) {
    println("ERR: " + e)
  }
}
```

```bash
$ golo golo --files data.golo lib.golo main.golo 
The size of [1, 2, 3] is 3!
The size of gololang.DynamicObject@4cb2c100 is 42!
ERR: java.lang.NoSuchMethodError: class java.lang.String::size
ERR: java.lang.NoSuchMethodError: class MyData.types.MyStruct::size
```

This issue is connected to #235

yloiseau:
When an augmentation is defined, it is only visible if the module using
it has imported the module defining it. This limits the ability to
create generic functions using any augmented object.

The solution is to look up in the call stack if a complying augmentation
had been imported in the calling module.

### DIFF
--- a/doc/augmentations.asciidoc
+++ b/doc/augmentations.asciidoc
@@ -131,6 +131,19 @@ TIP: As a matter of style, we suggest that your module names end with `Augmentat
 module imports **all** of its augmentation definitions, we suggest that you modularize them with *fine
 taste* (for what it means).
 
+To allow building libraries of functions leveraging augmentations, the callstack is also looked up in order to find if a complying augmentation was applied to the function arguments. It is thus possible to create a function like:
+
+[source,golo]
+----
+module MyLib
+
+function sayPlop = |o| {
+  println(o: plop())
+}
+----
+
+that can be applied on any object having a `plop` method, provided this is a native method or it is defined in an augmentation imported in the calling module.
+
 
 === Named augmentations
 
--- a/src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
+++ b/src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
@@ -11,6 +11,9 @@ package org.eclipse.golo.runtime;
 
 import java.lang.invoke.*;
 import java.lang.reflect.Method;
+import java.util.List;
+import static java.util.stream.Collectors.toList;
+import static java.util.Arrays.asList;
 
 import static java.lang.invoke.MethodHandles.*;
 import static java.lang.reflect.Modifier.*;
@@ -36,7 +39,7 @@ class AugmentationMethodFinder implements MethodFinder {
   };
   private final String[] argumentNames;
 
-  public AugmentationMethodFinder(MethodInvocationSupport.InlineCache inlineCache, Class<?> receiverClass, Object[] args) {
+  AugmentationMethodFinder(MethodInvocationSupport.InlineCache inlineCache, Class<?> receiverClass, Object[] args) {
     this.receiverClass = receiverClass;
     this.args = args;
     this.methodName = inlineCache.name;
@@ -48,7 +51,7 @@ class AugmentationMethodFinder implements MethodFinder {
     this.methodHandle = null;
     this.argumentNames = new String[inlineCache.argumentNames.length + 1];
     this.argumentNames[0] = "this";
-    System.arraycopy(inlineCache.argumentNames,0, argumentNames, 1, inlineCache.argumentNames.length);
+    System.arraycopy(inlineCache.argumentNames, 0, argumentNames, 1, inlineCache.argumentNames.length);
   }
 
 
@@ -112,7 +115,9 @@ class AugmentationMethodFinder implements MethodFinder {
       MethodHandle method;
       for (String augmentationName : Module.augmentationApplications(definingModule, augmentedClass)) {
         method = findMethod(augmentationClassName(definingModule, augmentationName));
-        if (method != null) { return method; }
+        if (method != null) {
+          return method;
+        }
       }
       return null;
     }
@@ -129,7 +134,7 @@ class AugmentationMethodFinder implements MethodFinder {
     protected String augmentationClassName(Class<?> definingModule, String augmentationName) {
       int idx = augmentationName.lastIndexOf(".");
       if (idx == -1) { return augmentationName; }
-      return (new StringBuilder(augmentationName)).replace(idx, idx+1, "$").toString();
+      return (new StringBuilder(augmentationName)).replace(idx, idx + 1, "$").toString();
     }
   }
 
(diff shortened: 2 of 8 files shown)


doc/augmentations.asciidoc
```diff
@@ -131,6 +131,19 @@ TIP: As a matter of style, we suggest that your module names end with `Augmentat
 module imports **all** of its augmentation definitions, we suggest that you modularize them with *fine
 taste* (for what it means).
 
+To allow to build libraries of functions leveraging augmentations, the callstack is also looked up in order to find if a complying augmentation was applied to the function arguments. It is thus possible to create a function like
```
jponge: "To allow building" (,,,)
(,,,) "a function like:"

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -36,7 +39,7 @@
   };
   private final String[] argumentNames;
 
-  public AugmentationMethodFinder(MethodInvocationSupport.InlineCache inlineCache, Class<?> receiverClass, Object[] args) {
+  AugmentationMethodFinder(MethodInvocationSupport.InlineCache inlineCache, Class<?> receiverClass, Object[] args) {
```
jponge: Why switching from `public` to _package-protected_?

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -129,7 +132,7 @@ public MethodHandle find(Class<?> definingModule, Class<?> augmentedClass) {
     protected String augmentationClassName(Class<?> definingModule, String augmentationName) {
       int idx = augmentationName.lastIndexOf(".");
       if (idx == -1) { return augmentationName; }
-      return (new StringBuilder(augmentationName)).replace(idx, idx+1, "$").toString();
+      return (new StringBuilder(augmentationName)).replace(idx, idx + 1, "$").toString();
```
jponge: I wonder if it is a good idea to use `StringBuilder` instead of `String`. Not a big deal, though.

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -221,17 +223,41 @@ private void findInImportedClasses(Class<?> sourceClass, FindingStrategy strateg
         Class<?> importedClass = classLoader.loadClass(importSymbol);
         findInClasses(importedClass, strategy);
         if (methodHandle != null) { break; }
-      } catch (ClassNotFoundException ignored) {}
+      } catch (ClassNotFoundException ignored) { }
     }
   }
 
+  private void findInCallStackImports(Class<?> callerClass, FindingStrategy strategy) {
+    if (methodHandle != null) { return; }
```
jponge: As a matter of style:

```java
if (cond) {
  break / return / anything
}
```

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -221,17 +223,41 @@ private void findInImportedClasses(Class<?> sourceClass, FindingStrategy strateg
...
+      try {
+        Class<?> callingClass = classLoader.loadClass(calling);
+        findInClasses(callingClass, strategy);
+        findInImportedClasses(callingClass, strategy);
+        if (methodHandle != null) { break; }
+      } catch (ClassNotFoundException ignored) { }
+    }
+  }
+
+  private List<String> getCallStack() {
```
jponge: :clap:

jponge:
Also a potential fix for #235

jponge:
Looks good to me.

I think @k33g could tell us wether this fixes #235 or not.

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
yloiseau: Because the class is not public anyway...

jponge: Huuu.... right :smiley:

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
yloiseau: AFAIK, `String::replace` can't take replace a given index, it's all or nothing

jponge: But since you know the index, you can replace with a string concatenation, which ends up being a `StringBuilder`. So yes, it’s better with `StringBuilder` as you did.

Note that JDK9 will use `invokedynamic` on some string concatenation call sites, to dispatch to optimized variants. Weird and funny stuff, isn’t it? :smile:

See http://openjdk.java.net/jeps/280

yloiseau: amazing :smile:

yloiseau:
Ok for me... BTW, this code will soon be refactored to deal with the priorities we talked about in #326 :smile:

yloiseau:
It does *not* fix #235. I think it's more a class loader issue.

jponge:
Regarding #235: there is perhaps a classloader issue as you say, but still it looks like having the augmented definition in the call stack would help.

This is just a guess, I haven't checked :smiley:

yloiseau:
Did the path you proposed in #235 was applied ?

jponge:
No, the patch hasn't been applied. I think it should be, but I was waiting on @k33g feedback.

yloiseau:
Ok, I can try to apply it in this PR, as we are at it and it's a related issue.

jponge:
Thanks Yannick.

yloiseau:
The patch indeed fixed #235 :clap:

jponge:
@yloiseau That's great news :smiley:

Issue #198, opened 2014-11-07
Property-style methods


jponge:
Golo currently does no magic on method resolution. 

We _could_ offer a shorter property-style method syntax such that:

```golo
foo: getBar(): getBaz(): setThat(123)
```

could be written as:

```golo
foo: bar: baz: that(123)
```

danielpetisme:
shouldn't be closed? via #332

PR #332, opened 2015-12-22
Property-style method support

danielpetisme:
# Description
This PR adds a property support to Golo. If `the RegularMethodFinder` can't do
the resolution, then the `PropertyMethodFinder` will seek for an accessor or
mutator after ensuring the method name matches to a field name.
Related issue: #198 

# Todo
- [x] Manage booleans
- [x] Dynamic fluent API
- [x] Do not mask user-defined setter return value
- [x] Support computed fields.

# Question
* Is the fluent aspect a good thing? Is it to the language or to the dev to make the getter/setter fluent?

### DIFF
--- a/doc/java-interop.asciidoc
+++ b/doc/java-interop.asciidoc
@@ -271,6 +271,73 @@ println(foo: bar())
 ----
 ====
 
+=== Properties support
+
+Golo support property-style method calls.
+
+Given a property `name()`, Golo will translate the method call to a `{get|is or set}Name` method call. Obviously, if a field
+is not accessible (ie. `private`) and doesn't have a getter or setter, the resolution will fail.
+Finally, _write-only_ properties can be chained (ie: return the current `this` instance), unless a return value is defined in the according setter method.
+
+NOTE: The property resolution does not check if an according field exists. Basically all the `get|set|is|` methods are candidates to
+a property-style method invocation.
+
+[source,java]
+----
+public class Person {
+
+  private String name;
+  private boolean goloComitter;
+  private String email;
+  private int score;
+
+  public String getName() {
+  ...
+  }
+
+  public boolean isGoloComitter() {
+  ...
+  }
+
+  public void setName(String name) {
+    ...
+  }
+
+  public void setGoloCommitter(boolean goloCommitter) {
+  ...
+  }
+
+  public int setScore(int score) {
+    this.score = score;
+    return score;
+  }
+
+  public boolean isRockStar() {
+    return goloCommitter;
+  }
+}
+----
+====
+
+
+[source,golo]
+----
+let duke = Person()
+
+try {
+  duke: email("duke@golo-lang.org")
+} catch (e) {
+  require(e oftype java.lang.NoSuchMethodError.class, "the email field is private and no setter is defined")
+}
+
+require(duke: name("Duke"): goloCommiter(true) oftype Person.class, "the set mutators should be chained with fluent calls")
+require(duke: name() is "Duke", "should find the getName() accessor.")
+require(duke: goloCommitter() is true, "should invoke the isGoloCommitter accessor since the field it's a boolean")
+require(print(duke: rockStar() is true, "even if the field rockstar doesn't exists, it should invoke the isRockStar accessor")
+require(ducke: score(100) is 100, "setScore returns a value, the method isn't fluent")
+----
+====
+
 === Inner classes and enumerations
 
 We will illustrate both how to deal with public static inner classes and enumerations at once.
--- a/src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
+++ b/src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2012-2015 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.golo.runtime;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.reflect.Method;
+import java.util.Objects;
+import java.util.Optional;
+
+import static java.lang.invoke.MethodHandles.*;
+import static java.lang.invoke.MethodType.methodType;
+import static java.util.Arrays.copyOfRange;
+
+public class PropertyMethodFinder extends MethodFinder {
+
+  private static final MethodHandle FLUENT_SETTER;
+
+  static {
+    try {
+      FLUENT_SETTER = lookup().findStatic(
+          PropertyMethodFinder.class,
+          "fluentSetter",
+          methodType(Object.class, Object.class, Object.class));
+    } catch (NoSuchMethodException | IllegalAccessException e) {
+      throw new Error("Could not bootstrap the required fluent method handles", e);
+    }
+  }
+
+  private static Object fluentSetter(Object o, Object notUsedSetterReturnValue) {
+    return o;
+  }
+
+  private String propertyName;
+
+  public PropertyMethodFinder(MethodInvocation invocation, Lookup lookup) {
+    super(invocation, lookup);
+    this.propertyName = capitalize(invocation.name());
+  }
+
+  private static MethodInvocation propertyInvocation(String propertyInvocationName, MethodInvocation parentInvocation) {
+    return new MethodInvocation(
+        propertyInvocationName,
+        parentInvocation.type(),
+        parentInvocation.arguments(),
+        copyOfRange(parentInvocation.argumentNames(), 0, parentInvocation.argumentNames().length - 1));
+  }
+
+  private MethodHandle findMethodForGetter() {
+    MethodHandle target = new RegularMethodFinder(
+        propertyInvocation("get" + propertyName, invocation),
+        lookup
+    ).find();
+
+    if (target != null) {
+      return target;
+    }
+    return new RegularMethodFinder(
+        propertyInvocation("is" + propertyName, invocation),
+        lookup
+    ).find();
+  }
+
+  private MethodHandle fluentMethodHandle(Method candidate) {
+    Objects.requireNonNull(candidate);
+    MethodHandle target = toMethodHandle(candidate).orElse(null);
+    if (target != null) {
+      if (!TypeMatching.returnsValue(candidate)) {
+        Object receiver = invocation.arguments()[0];
+        MethodHandle fluent = FLUENT_SETTER.bindTo(receiver);
+        target = filterReturnValue(target, fluent);
+      }
+    }
+    return target;
+  }
+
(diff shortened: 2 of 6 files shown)


yloiseau:
If by “fluent” you just mean that the setter returns `this`, I think it's ok if the original setter returns `void` (which *should* be the case, but better check it). Otherwise, we must obviously not change the returned value.

jponge:
Looks like a worthy feature to add.

I concur with @Artpej on conventions for `boolean` properties, `isBlaBlaBla()` is often preferred to `getBlaBlaBla()`.

On _fluency_: I tend to like that, especially when most of the time it boils down to simply returning `this`. That being said @yloiseau is right: this should only be done unless there exists a non-`void` `setFoo(...)` method.

danielpetisme:
Boolean OK.

So far the properties support is only a hack on the method resolution (at runtime). Fluency implies to add bytecode instructions to setters to return the current object. This has to be made during the compilation stage. My question is how to detect a property call (vs a standard method or field call) during the compilation.

jponge:
You can't, it's a dynamically typed language so everything happens at runtime.

yloiseau:
The wrapping of the method handle at resolution time is the way to go IMO

yloiseau:
Maybe something as simple as `filterReturnValue(handle, constant(receiver)`

jponge:
+1

danielpetisme:
Done!
I declared a `fluent` method which basically return the  current object. Then, if the target method is a setter, I compose it with the `fluent` method thanks to `MethodHandles#filterReturnValue`. 
And voilà ;)

danielpetisme:
Ready for review

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private static Object fluentSetter(Object o, Object notUsedSetterReturnValue) {
+    return o;
+  }
+
+  public PropertyMethodFinder(MethodInvocationSupport.InlineCache inlineCache, Class<?> receiverClass, Object[] args) {
+    super(inlineCache, receiverClass, args);
+  }
+
+  private MethodHandle fluentSetterMethodHandle() {
```
yloiseau: Maybe this could be a static final field with static initialization, as the FALLBACK in `*Support`

yloiseau:
Note on boolean properties: when doing a “fluent” interface by hand, I personally tends to create a parameterless setter `fooable()`, the property beeing initialized to `false` (or a setter `notFooable` with initialization to `true`) and a getter named `isFooable()`, that is:
```java
class Person {
    private boolean isAwesome = false;
    public boolean isAwesome() { return this.isAwesome; }
    public Person awesome() { this.isAwesome = true; return this; }    
}

Person ford = new Person().firstname("Ford").name("Perfect").awesome();
if (ford.isAwesome()) { ... }
```

instead of 
```java
class Person {
    private boolean isAwesome;
    public boolean awesome() { return this.isAwesome; }
    public Person awesome(boolean v) { this.isAwesome = v; return this; }    
}

Person ford = new Person().firstname("Ford").name("Perfect").awesome(true);
if (ford.awesome()) { ... }
```

I find it more readable, but it's a matter of taste. For instance `mylist.empty()` sound like I want to clear the list to me, while `mylist.isEmpty` is clearer.

 Moreover, it would be (a little) more complicated and (maybe) more risky to generate such methods.

danielpetisme:
Regarding setters return value

A quick, and quite naive, solution to not mask the user-defined setter return value, is to check if this value is null or not. Indeed, when a method return value is `void`, in fact it returns `null`. But what if the actual setter return value is null (let's say the user hard coded it). This implementation will detect it as a `void` setter.
So, not good.
Internally, `RegularMethodFinder` target method signature is represented by generic types (ie. `Object`). So we can rely on `RegularMethodFinder#type` to know if the setter use `void` or not.

The ideal solution is to get the actual `Method` found by `RegularMethodFinder#findInMethods` before its conversion to MethodHandle. With the actual `Method` object we can easily check the return type.
This solution implies a `RegularMethodFinder` refactoring as @yloiseau is doing in #334.
So I propose to pause the PR until #334 is merged.

danielpetisme:
Open for review.

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
```diff
@@ -0,0 +1,103 @@
...
+  private static Object fluentSetter(Object o, Object notUsedSetterReturnValue) {
+    return o;
+  }
+
+  public PropertyMethodFinder(MethodInvocation invocation, Lookup lookup) {
+    super(invocation, lookup);
+  }
+
+  private MethodHandle fluentSetterMethodHandle() {
+    MethodHandle mh;
```
yloiseau: can't you make this method handle a static field?

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.lang.invoke.MethodHandle;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+import static java.lang.invoke.MethodHandles.*;
+import static java.lang.invoke.MethodType.methodType;
+import static java.util.Arrays.copyOfRange;
+
+public class PropertyMethodFinder extends RegularMethodFinder {
```
yloiseau: Why do you extends `RegularMethodFinder`? Looks like you only delegate. Maybe extending `MethodFinder` would be enough.

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
```diff
@@ -0,0 +1,103 @@
...
+      return findMethodForGetter(field);
+    }
+    return findMethodForSetter(field);
+  }
+
+  private boolean returnsValue(Method m) {
+    return !(m.getReturnType().equals(void.class) || m.getReturnType().equals(Void.class));
+  }
+
+  private String capitalize(String word) {
```
yloiseau: Can be static

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
danielpetisme: I use `RegularMethodFinder#findInMethods` so I make the method protected and I extended it.
If don't see objections to make it public I can, indeed, extend `MethodFinder` and delegate the property method research to `RMF`

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
```diff
@@ -0,0 +1,103 @@
...
+    } catch (NoSuchFieldException e) {
+      return null;
+    }
+    if (invocation.arguments().length == 1) {
+      return findMethodForGetter(field);
+    }
+    return findMethodForSetter(field);
+  }
+
+  private boolean returnsValue(Method m) {
```
yloiseau: Can be static

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Override
+  public MethodHandle find() {
+    Field field;
+    try {
+      field = invocation.receiverClass().getDeclaredField(invocation.name());
+    } catch (NoSuchFieldException e) {
+      return null;
+    }
+    if (invocation.arguments().length == 1) {
```
yloiseau: `invocation.arity()`

src/main/java/org/eclipse/golo/runtime/PropertyMethodFinder.java
yloiseau: I don't see where you call the inherited method... :smile:

danielpetisme: Line 66

yloiseau: You call it on a `RegularMethodFinder`... Ok, got it, it's just a access problem, right? In that case, I rather make it public and extend `MethodFinder`

yloiseau:
Given
```java
class Person {
  private Date birthday;
  
  public Date getBirthday() { return birthday; }

  public int getAge() { ... }
}
```
we can't do
```golo
let p = Person(...)
p: age()
```
since you abort the search if the class does not have a real field, while
```golo
p: getAge()
```
will work.

I feel it counter intuitive. It breaks encapsulation, since the user of the class should not have to know if `age` is a real field or a computed one.

danielpetisme:
It's a valid point. Frankly, I was lazy so I just followed the classical [JavaBeans properties conventions](http://docs.oracle.com/javase/tutorial/javabeans/writing/properties.html).

If the skip the field constraint in that case every method starting with `is|get|set` is candidate to this feature even if it has nothing to do with the class fields (it would more act as a sugar than a property).

I'm open on this point.

jponge:
I think resolving based on method conventions is great, and that we shouldn't care about fields. Otherwise we would be leaking internals, although we all know many `get`/`set` methods just do that :wink: 

But still, it's how an object-oriented design should work: public accessors may or may not be directly backing a field, hence I concur with @yloiseau

danielpetisme:
Updated with @yloiseau + @jponge remarks.
Ready for another round of review.

doc/java-interop.asciidoc
```diff
@@ -271,6 +271,74 @@ println(foo: bar())
 ----
 ====
 
+=== Properties support
+
+Golo support property-style method calls.
+
+Given a property `name()`, Golo try to find a public instance field named `name` or a method `name()`. If the resolution fails,
+it will search for a `{get|is or set}Name` method. Obviously, if a field is not accessible (ie. `private`) and doesn't have a getter
```
jponge: (...) it will *now* (...)

doc/java-interop.asciidoc
```diff
@@ -271,6 +271,74 @@ println(foo: bar())
...
+or setter, the resolution will also fail. Finally, because fluency rocks, the setter methods return the current instance (ie: `this`) unless a return value is
```
jponge: Style: _fluency rocks_ does not mean anything :-)

doc/java-interop.asciidoc
```diff
@@ -271,6 +271,74 @@ println(foo: bar())
...
+
+  public boolean isRockStar() {
+    return goloCommitter;
+  }
+}
+----
+====
+
+
+[source,golo]
```
jponge: :+1:

jponge:
I don't have any further comment except those on the documentation. Nice piece of work @danielpetisme :beers:

danielpetisme:
@jponge I tried to be more explicit on the `fluent` aspect, I hope it's clear enough.
Let me know if the review stage is done, I'll squash the commits before the final merge.

It was an interesting PR to do ;) Easy to do code but tricky in the corner cases.

jponge:
Good for me!

You can avoid squashing commits if you like.

danielpetisme:
Commits squashed, ready to merge (still don't have the permission & yes I know I have to raise a ticket to Eclipse... :disappointed:)

jponge:
:+1:

jponge:
:+1:

PR #334, opened 2015-12-23
Refactor method and augmentation resolution

### LINKED ISSUE #275: Problem with varargs augmentation
The following Golo code:
```golo
struct Test = { v }

augment Test {
  function foo = |this, args...| -> "## Hey: " + this: v(): toString() + args: asList(): toString()
}

function main = |args| {
  let f = Test(42)
  println(f: foo("a"))
  println(f: foo())
}
```
gives:
```
## Hey: 42[a]
Exception in thread "main" java.lang.NoSuchMethodError: class pb.types.Test::foo
	at fr.insalyon.citi.golo.runtime.MethodInvocationSupport.findTarget(MethodInvocationSupport.java:217)
	at fr.insalyon.citi.golo.runtime.MethodInvocationSupport.lookupTarget(MethodInvocationSupport.java:141)
...
```

Probably a bug when dealing with varargs.

yloiseau:
Change the way (augmented) methods are resolved at runtime, to deal
with the priorities discussed in #326.

Move methods common between augmentation and regular method resolution
into a abstract superclass.

Refactor method type matching, to better reuse and move it to the
MethodInvocation object.

FIX #275

Also some refactoring on tests.

FIX: empty augmentation

Not filled as an issue, but a class was generated to hold augment
functions, even if there was no functions. This happened when using
named augmentation, since a single IR node is created for declarations
such as

    augment Foo {
      function plop = |this| -> "plop"
    }

and

    augment Foo with Plop

In the first case a class `module$Foo` is created to hold the `plop`
function. It was also created in the second case, but was empty.


build.gradle
```diff
@@ -130,6 +130,16 @@ tasks.withType(FindBugs) {
   }
 }
 
+jacocoTestReport {
```
jponge: By default the `jacoco` plugin does generate HTML. Is there a particular reason for making that plugin configuration explicit?

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -10,282 +10,72 @@
...
+import java.util.Comparator;
+import java.util.Objects;
+import org.eclipse.golo.runtime.augmentation.DefiningModule;
 
 import static java.lang.invoke.MethodHandles.*;
 import static java.lang.reflect.Modifier.*;
-import static org.eclipse.golo.runtime.DecoratorsHelper.getDecoratedMethodHandle;
-import static org.eclipse.golo.runtime.DecoratorsHelper.isMethodDecorated;
 
-class AugmentationMethodFinder implements MethodFinder {
```
jponge: Ok, so `MethodInvocation` is a domain object to describe a particular lookup. Fine refactoring :smile:

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -10,282 +10,72 @@
...
-      try {
-        Class<?> callingClass = classLoader.loadClass(calling);
-        findInClasses(callingClass, strategy);
-        findInImportedClasses(callingClass, strategy);
-        if (methodHandle != null) {
-          break;
-        }
-      } catch (ClassNotFoundException ignored) { }
-    }
+  private static boolean candidateStackTrace(String className) {
```
jponge: Perhaps `isCandidateInStackTrace`?

src/main/java/org/eclipse/golo/runtime/RegularMethodFinder.java
```diff
@@ -113,69 +88,21 @@ private static String reverseStructAugmentation(String receiverClassName) {
...
-      Class<?>[] parameterTypes = method.getParameterTypes();
-      Object[] argsWithoutReceiver = copyOfRange(args, 1, args.length);
-      if (TypeMatching.argumentsNumberMatch(argsWithoutReceiver, method, parameterTypes)) {
-        if (TypeMatching.canAssign(parameterTypes, argsWithoutReceiver, method.isVarArgs())) {
-          return toMethodHandle(method);
-        }
-      }
-    }
-    return null;
+  private Stream<Optional<MethodHandle>> findInMethods() {
```
danielpetisme: I propose to return `Method` instead of `MethodHandle` see [my comment](https://github.com/eclipse/golo-lang/pull/332#issuecomment-166919207) on #332.

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
```diff
@@ -10,282 +10,72 @@
...
-      findInClasses(callerClass, strategy);
-    }
-    for (FindingStrategy strategy : strategies) {
-      findInImportedClasses(callerClass, strategy);
-    }
-    for (FindingStrategy strategy : strategies) {
-      findInCallStackImports(callerClass, strategy);
-    }
-    return methodHandle;
+    return getDefiningModules()
```
jponge: I can picture your smile @yloiseau while you typed this.

src/main/java/org/eclipse/golo/runtime/Loader.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package org.eclipse.golo.runtime;
+
+import java.util.function.Function;
+
+/**
+ * Encapsulate a {@code ClassLoader} to ease streaming.
+ */
+public final class Loader implements Function<String, Class<?>> {
```
jponge: You can probably complete the javadocs.

src/main/java/org/eclipse/golo/runtime/MethodInvocation.java
```diff
@@ -0,0 +1,61 @@
...
+import java.lang.invoke.MethodType;
+import java.lang.invoke.MethodHandle;
+import java.lang.reflect.Method;
+
+import static java.lang.reflect.Modifier.*;
+
+/**
+ * Encapsulate informations about a runtime method call.
+ */
+public class MethodInvocation {
```
jponge: Does it have to be public?

src/main/java/org/eclipse/golo/runtime/MethodInvocationSupport.java
```diff
@@ -20,7 +20,11 @@
 import static java.lang.invoke.MethodType.methodType;
 
 
-public class MethodInvocationSupport {
+public final class MethodInvocationSupport {
+
+  private MethodInvocationSupport() {
+    throw new UnsupportedOperationException("utility class");
```
jponge: I wouldn't call classes with `invokedynamic` bootstrap methods "utility classes" :smile:

src/main/java/org/eclipse/golo/runtime/Module.java
```diff
@@ -12,16 +12,20 @@
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
-class Module {
+public final class Module {
```
jponge: Same here, if we can lower visibility then it would be better.

build.gradle
yloiseau: It does not seems to generate anything here.

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
yloiseau: :smile:

src/main/java/org/eclipse/golo/runtime/Module.java
yloiseau: It's now used in a subpackage (`runtime.augmentation)`, and thus need to be public. Either that, or move up the objects from `runtime.augmentation`, but I didn't want to clutter the `runtime` package. Both have drawback.

src/main/java/org/eclipse/golo/runtime/AugmentationMethodFinder.java
yloiseau: Thanks :smile:

src/main/java/org/eclipse/golo/runtime/augmentation/package-info.java
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+
+/**
+ * Helper class to resolve methods added by an augmentation.
```
jponge: class*es*

build.gradle
jponge: There is task for that: `./gradlew jacocoTestReport` (it reads the JaCoCo-produced binary file and outputs the nice HTML reports).

jponge: The behaviour by default is to generate a compact report file, and call the `jacocoTestReport` task when you want a fancy HTML report. This also saves a tiny amount of build time.

src/main/java/org/eclipse/golo/runtime/Module.java
jponge: Ah yes, ok then if you go with a subpackage.

jponge:
An now the `ip-validation` check does not like your correct email address. Same issue I had the other day :walking:

build.gradle
yloiseau: Ok, I know what my problem was. The `jacocoTest` or `jacocoTestReport` tasks don't do anything  if no tests were run. 
```
./gradlew --daemon jacocoTest
[...]
:jacocoTestReport SKIPPED
```
I supposed their was a dependency between the two.  Adding it to `check` dependencies solved “my issue” since `check` depends on tests.

src/main/java/org/eclipse/golo/runtime/MethodInvocation.java
yloiseau: Used in subpackage...

build.gradle
```diff
@@ -130,6 +130,9 @@ tasks.withType(FindBugs) {
   }
 }
 
+check.dependsOn jacocoTestReport
```
jponge: I'd still remove it.

The other line on `jacocoTestReport` depending on `test` is ok.

jponge:
@yloiseau good for me, modulo the point on `jacocoTestReport`

jponge:
@yloiseau Just in case you have time, a boolean method in `RegularMethodFinder` to check if we are in presence of a Java overloaded method would help (for the method name, several methods have the same number of arguments).

I've started a branch to allow dealing with such methods, as in:

```java
/*
 * Copyright (c) 2012-2015 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
 *
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 */

package org.eclipse.golo.compiler.testing.support;

public class ClassWithOverloadedMethods {

  public String foo(String str) {
    return "# " + str;
  }

  public String foo(int i) {
    return "% " + i;
  }
}
```

I made a first commit with a unit test, and while I looked into `RegularMethodFinder` and `MethodInvocationSupport`, I thought that I'd better wait for that PR to be merged.

BTW I think the strategy for overloaded methods will be a simple inline-cache with a `guardWithTest` that checks for receiver *and* argument types. For performance reasons I'll probably make several arity guard targets before falling back to an array boxing one above a certain number of arguments).

yloiseau:
@jponge  unless I missed the point, I think we already have the correct behavior. Given:
```java
package test;

public class OverloadedMethod {
  public String foo(String s) {
    return "String: " + s;
  }

  public String foo(Integer i) {
    return "Integer: " + i;
  }

  public String foo(Number n) {
    return "Number: " + n;
  }
}
```
and 
```golo
module TestOverloaded

import test

function main = |args| {
  let t = OverloadedMethod()
  println(t: foo("a"))
  println(t: foo(1))
  println(t: foo(2.5))
}
```

```
$ javac -d . OverloadedMethod.java && golo golo --classpath . --files overloaded.golo
String: a
Number: 1
Number: 2.5
```

yloiseau:
oops nop... should be `Integer: 1`... grrrr

jponge:
The first call wins :-)

I just need the resolution algorithm to tell me if it was an overload and then I'll deal with the call site plumbing logic.

jponge:
I should have been more precise: _"the first call wins"_ is rather when you have, say, a loop:

```golo
let f = SomeObject()
let a = [1, "plop", true]
foreach i in a {
  f: plop(i)
}
```

(with the assumption that `SomeObject` is a constructor to some Java class where `foo` is overloaded for the 3 different types.

yloiseau:
Indeed, and the one chosen is the “first defined” one… by swapping the java methods, we get other behavior. 
BTW, won't you also need the original method signature?

jponge:
My understanding is that I just need to know that we have a call site with overloaded methods. I don't think the signature is useful.

The idea is to have a special behavior for these call sites, done with simple inline caches. Invalidation will trigger a new lookup, which must resolve fine.

yloiseau:
Something like this? Maybe we could change the sorting criteria in `Extractors::getMethods` to sort the more specific ones first...

jponge:
@yloiseau See my comments, but just a simple flag as suggested is enough. The idea is that if the resolution tells me we had overloads, then I'll do something different at the call site level:

1. reset and trash any previously cached target, and
2. make it a simple inline-cache, and
3. mount a MH guard that also checks parameter types, and invalidates if a change is spotted (e.g. you had resolved to `foo(Integer)` and now you pass a `String`, so you must invalidate the call site).

jponge:
BTW this will make such call sites slower, but this will be a feature to improve interop with Java. I think it will solve most issues we had with heavily overloaded method APIs, e.g., the ones from Vert.x.

jponge:
Looks good to me!

yloiseau:
Ok, let's merge

jponge:
And so you close the 2015 year of Golo :tada:

yloiseau:
it's a privilege :smile:

PR #337, opened 2015-12-28
Fix cursor movement

danielpetisme:
Replace `println` by `print` to not jump to the new line when the cursor is moved.


jponge:
Thanks!

PR #339, opened 2016-01-02
Support Java overloaded instance methods

jponge:
This supports resolving and invalidating Java overloaded instance methods.

### DIFF
--- a/src/main/java/org/eclipse/golo/runtime/RegularMethodFinder.java
+++ b/src/main/java/org/eclipse/golo/runtime/RegularMethodFinder.java
@@ -9,18 +9,18 @@
 
 package org.eclipse.golo.runtime;
 
-import java.lang.invoke.*;
-import java.lang.reflect.Field;
-import java.lang.reflect.Method;
-import java.util.stream.Stream;
-import java.util.Optional;
-import java.util.List;
 import gololang.GoloStruct;
 
-import static java.util.stream.Collectors.toList;
-import static java.util.Collections.unmodifiableList;
+import java.lang.invoke.MethodHandle;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
 import static java.lang.invoke.MethodHandles.*;
 import static java.lang.reflect.Modifier.*;
+import static java.util.stream.Collectors.toList;
 
 class RegularMethodFinder extends MethodFinder {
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/eclipse/golo/runtime/RegularMethodFinder.java
```diff
@@ -9,44 +9,48 @@
...
+    return Extractors.getMethods(invocation.receiverClass())
+        .filter(this::overloadMatch)
+        .count() > 1;
+  }
+
+  private boolean overloadMatch(Method m) {
+    return Extractors.isPublic(m) &&
+        Extractors.isConcrete(m) &&
+        m.getName().equals(invocation.name()) &&
+        (m.getParameterCount() + 1 == invocation.arity()) || (m.isVarArgs() && (m.getParameterCount() <= invocation.arity()));
```
yloiseau: looks like déjà vu... maybe `TypeMatching.argumentsNumberMatches` ?

src/main/java/org/eclipse/golo/runtime/RegularMethodFinder.java
```diff
@@ -82,37 +86,39 @@ public boolean isOverloaded() {
...
   }
 
   private static String reverseStructAugmentation(String receiverClassName) {
     return receiverClassName.substring(0, receiverClassName.indexOf(".types"))
-      + "$" + receiverClassName.replace('.', '$');
+        + "$" + receiverClassName.replace('.', '$');
   }
 
 
   private Stream<Method> findInMethods() {
```
danielpetisme: Is it meaningful to extract `findInMethods`, `findInFields`, `toMethodHandles` and others into a Reflect/InvokeUtils ?

yloiseau: We indeed have a code duplication (e.g. in `FunctionCallSupport` and `Predefined::fun`), but AFAIR, these particular methods are specific to a particular finder, so I don't think we can really extract common code. Type matching, checking and filtering are also candidates to possible duplication (even if I tried to remove as much as possible).
BTW, I started to extract the naming logic into a separate utility class, but didn't integrated it in my previous refactoring PR. Indeed, we are doing special names manipulations (like `MyModule.types.MyStruct`, augmentations replacing `.` with `$`, etc.) in several separated places: IR construction, bytecode generation, runtime lookup to name a few.

src/main/java/org/eclipse/golo/runtime/RegularMethodFinder.java
jponge: I need to try it :-)

jponge: No, it doesn't fit.

jponge:
@yloiseau This makes me think that `FunctionCallSupport` is a good refactoring candidate.

This PR focuses only on instance methods, but once refactored a bit we could improve `FunctionCallSupport` to also deal with overloaded Java static methods.

jponge:
Recorded #341 for overloaded static methods.

I'm ok with this PR, if you agree then I'll go ahead and merge.

yloiseau:
:+1:

PR #343, opened 2016-01-06
Make the orIfNull operator lazy

### LINKED ISSUE #342: orIfNull is eager
The `orIfNull` operator is currently eager, that is the right argument is evaluated even if the left one is not null. For instance
```golo
function plop = {
  println("da plop")
  return 42
}

function main = |args| {
  let a = 69 orIfNull plop()
}
```
will print `da plop`.

It would be more useful if `orIfNull` were lazy, like boolean operators.

yloiseau:
FIX #342

Replace the call to an invoke dynamic method by a generated bytecode
whose evaluation is lazy. For instance, in:
```golo
"plop" orIfNull getPlop()
```
the `getPlop` function is not called anymore.


jponge:
Excellent!

PR #345, opened 2016-01-07
FIX #323: Check for duplicated types compile time

### LINKED ISSUE #323: Duplicated struct not identified at compile time
When a module contains two identically named structures, as in
```golo
module Test

struct Toto = { x }
  
struct Toto = { x, y }

function main = |args| {
  let t2 = Toto(1, 2)
}
```
the code compile without complaining, but fail at runtime, with `java.lang.ClassFormatError: Duplicate method name&signature` when `golo run` and with  `java.lang.LinkageError: loader (instance of GoloClassLoader): attempted duplicate class definition` when `golo golo`.

*Intended behavior*: should fail at *compile time* with explicit error, e.g.
```
[err] Duplicated `struct` definition  for "Toto" at line 5, column 1
```

yloiseau:
When two types (`struct` or `union`) with the same name were declared in
the same module, the code compiled without error, but an obscure
exception was raised at runtime

The case is now checked when building the IR, so that we have a clear
compiling error.

This fix the #323 issue.

### DIFF
--- a/src/main/java/org/eclipse/golo/compiler/ir/GoloModule.java
+++ b/src/main/java/org/eclipse/golo/compiler/ir/GoloModule.java
@@ -135,6 +135,20 @@ public final class GoloModule extends GoloElement implements FunctionContainer {
     struct.setModuleName(getPackageAndClass());
   }
 
+  public GoloElement getSubtypeByName(String name) {
+    for (Struct s : structs) {
+      if (s.getName().equals(name)) {
+        return s;
+      }
+    }
+    for (Union u : unions) {
+      if (u.getName().equals(name)) {
+        return u;
+      }
+    }
+    return null;
+  }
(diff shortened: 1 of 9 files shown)


src/main/java/org/eclipse/golo/compiler/ir/GoloModule.java
```diff
@@ -135,6 +135,20 @@ public void addStruct(Struct struct) {
     struct.setModuleName(getPackageAndClass());
   }
 
+  public GoloElement getSubtypeByName(String name) {
```
jponge: I'm surprised you did not use the Java stream API :trollface: 

Anyway you are right, this exists early without creating intermediate objects :+1:

jponge:
Looks good to me!

Issue #192, opened 2014-10-21
Macros


jponge:
Wouldn't it be fun to have (hygienic) macros? :smile: 

https://en.wikipedia.org/wiki/Hygienic_macro

jponge:
I think this is done 😉

PR #347, opened 2016-01-09
Hygienic macros

yloiseau:
This PR adds hygienic macros to the golo language (#192) :tada:

It's not ready yet, but I make the PR for feedbacks.

### TODO
- [x] more tests
- [x] comprehensive and comprehensible documentation
- [x] update `golo doc` do deal with macro definitions
- [x] better error reporting when a macro fails

### DIFF
--- a/src/main/java/org/eclipse/golo/compiler/macro/MacroCache.java
+++ b/src/main/java/org/eclipse/golo/compiler/macro/MacroCache.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2012-2018 Institut National des Sciences Appliquées de Lyon (INSA-Lyon) and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.golo.compiler.macro;
+
+import gololang.ir.AbstractInvocation;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+class MacroCache {
+  private final Map<MacroKey, MacroFinderResult> cache = new HashMap<>();
+
+  private static final class MacroKey {
+    private final String name;
+    private final int arity;
+
+    MacroKey(String name, int arity) {
+      this.name = name;
+      this.arity = arity;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (o == null) { return false; }
+      if (o == this) { return true; }
+      if (!(o instanceof MacroKey)) { return false; }
+      MacroKey that = (MacroKey) o;
+      return this.name.equals(that.name) && this.arity == that.arity;
+    }
+
+    @Override
+    public int hashCode() {
+      return java.util.Objects.hash(name, arity);
+    }
+
+    public static MacroKey of(AbstractInvocation<?> invocation) {
+      return new MacroKey(invocation.getName(), invocation.getArity());
+    }
+
+    @Override
+    public String toString() {
+      return String.format("MacroKey<%s,%s>", name, arity);
+    }
+  }
+
+  public void clear() {
+    cache.clear();
+  }
+
+  public Optional<MacroFinderResult> getOrCompute(AbstractInvocation<?> invocation, Function<AbstractInvocation<?>, Optional<MacroFinderResult>> finder) {
+    if (invocation == null) {
+      return Optional.empty();
+    }
+    MacroKey key = MacroKey.of(invocation);
+    if (cache.containsKey(key)) {
+      return Optional.of(cache.get(key));
+    }
+    Optional<MacroFinderResult> method = finder.apply(invocation);
+    if (method.isPresent()) {
+      cache.put(key, method.get());
+    }
+    return method;
+  }
+}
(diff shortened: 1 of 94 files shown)


jponge:
Will check later :-)

src/main/java/org/eclipse/golo/cli/command/ClasspathOption.java
```diff
@@ -0,0 +1,52 @@
...
+class ClasspathOption {
+
+  @Parameter(names = "--classpath", variableArity = true, description = "Classpath elements (.jar and directories)")
+  List<String> classpath = new LinkedList<>();
+
+  URLClassLoader primaryClassLoader() throws MalformedURLException {
+    classpath.add(0, ".");
+    URL[] urls = new URL[classpath.size()];
+    int index = 0;
+    for (String element : classpath) {
```
jponge: Just throw a `for` loop with an `int` / guard / increment :smile_cat:

src/main/java/org/eclipse/golo/compiler/macro/MacroCache.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Map;
+import java.util.HashMap;
+import java.lang.invoke.MethodHandle;
+import java.util.function.Function;
+
+import org.eclipse.golo.compiler.ir.MacroInvocation;
+
+class MacroCache {
+  private final Map<MacroKey, MethodHandle> cache = new HashMap<>();
```
jponge: How about a `WeakHashMap`?

jponge:
This is promising!

src/main/java/org/eclipse/golo/compiler/macro/MacroCache.java
yloiseau: Since it's just a compile time cache, I'm not sure it worth the blow

jponge: Oh yes

yloiseau:
I'm not sure of the "contextual macro" stuff. Briefly, it's a way to implicitly call a macro with its context as first parameter. The "context" is the enclosing node of some kind (case, loop, function definition, etc.)
The idea is to allow the macro to modify its context. 
However, I wonder if the macro call itself as first argument would be more useful, since it would let the macro code to navigate the tree more explicitly. I fear that the notion on context be not explicit enough. Note that one can navigate from the macro argument, but this would be impossible in the case of parameterless macros. What do you think?

jponge:
@yloiseau explicit as a first argument is better I think

yloiseau:
@jponge would you mind to develop? The current implementation provides a special call syntax `&:theMacro(args)` that call the macro with its context as first argument, the context being the first *macro call* parent node of some kind. I have two concerns here: the special syntax and the context concept. I see two alternatives: keep a special syntax, but use the macro call itself as first argument, and use navigation methods to retrieve the desired node (e.g. `getAncestorOfType`), maybe with flagging interfaces on some golo elements; or remove the special syntax and make any macro call have itself as first argument. I don't see how we could call the macro giving a parent node explicitly.

jponge:
Do you have a sample that you could paste?
Le jeu. 17 août 2017 à 12:42, Yannick Loiseau <notifications@github.com> a
écrit :

yloiseau:
For instance:
```golo
macro importUsualModules = |mod| {
  require(mod oftype GoloModule.class, "This macro must be contextually called toplevel")
  foreach name in list[...] {
    mod: addImport(moduleImport(name))
  }
}
```

which adds some predefined imports into the given module.
It can be used as:
```golo
module MyModule

# ...

&:importUsualModules()

# ...
```
since in that call (toplevel), the context is the module itself. If the macro had another parameter (`arg`), we could have drop the context and use `arg: enclosingModule(): addImport(...)`, but here, we have no node to query for its parent, and we can't pass explicitly the current module node to the macro.

My concern is that the context can be ambiguous, so I suggest that the implicit argument be the macro call itself. In this example, the macro becomes:
```golo
macro importUsualModules = |call| {
  require(call oftype MacroInvocation.class, "This macro must be called with a context")
  let mod = call: enclosingModule()
  foreach name in list[...] {
    mod: addImport(moduleImport(name))
  }
}
```

jponge:
Ok I'll test +  review, but give me a few days :-)

yloiseau:
@jponge no urgency, I'm still polishing some details. 

Concerning the context, I think I'll drop the notion on context completely, and use the macro call itself. It will be easier to gasp. If I recall correctly, I introduced it at a stage where I did not have added yet the IR navigation methods. One question remains:  should all macro calls have themselves as implicit parameters (à la sweetjs) or only the ones called with the special syntax (`&:`). I think the explicit syntax is better.

yloiseau:
It appears that in the current state, the `macro` keyword could be dropped and replaced by a macro itself.

Indeed, from the IR point of view, a macro is just a function (`o.e.g.c.ir.GoloFunction`) with a macro flag, wich is set by the parser if the function is declared using `macro` instead of `function`. It is thus trivial to create a (builtin) macro that takes a function and set the flag.

A macro would then be declared as

```golo
@macro
function myMacro = ...
```

instead of

```golo
macro myMacro = ...
```

removing the need for a new keyword.

What do you think?

utybo:
From a (beginner) developer standpoint, that would make one assume that a macro is just a special function, would that be ok?

yloiseau:
Yes, a macro is just a special function...

yloiseau:
ping @jponge @artpej @k33g

k33g:
@yloiseau I have some issues when compiling, but I'm on `openjdk 13.0.1`, do I need to use a lower jdk?

yloiseau:
Indeed, Golo is not java>9 ready yet... 😧

jponge:
Yes Golo needs to be compiled on Java 8. Most of the code base works fine on JDKs above, but the tests suite does not pass.

jponge:
Is this the longest open PR in history?

jponge:
Best.
Pull-request.
Ever.

👍

Issue #351, opened 2016-01-19
Compile with `javac -parameters`


jponge:
This `java` flag is useful for having parameter name metadata. This is what Golo uses for resolving named arguments.

This is not the case by default, so we should play a role model in supporting the feature for the source code of Golo.

danielpetisme:
This one is easy to fix...

IMHO, parameters names are useful in the APIs. if the language provide named argument the names are key and if not they still help the developer to understand what the function/method does.

In Golo, most of what the developer sees is already developed in Golo itself and natively mimic natively the `-parameters` flag. 
What I'm trying to say is, adding the `-parameters` option is fancy, we'll be (i think) the very first JVM-language to do it, but the real-benefits is very limited.

:+1:  to do it, just to be on the hype ;)

yloiseau:
Fixed by #352 as much as possible. The JDK itself is not compiled with `-parameters`, but nothing we can do about.

PR #346, opened 2016-01-07
FIX #335: Named arguments and anonymous calls

yloiseau:
There is a issue when doing anonymous calls, i.e. calls directly applied
on references, on lambda, on partialized functions, or on functions
returned by other functions, when using named arguments.

The issue is due to an incorrect flagging in the AST to IR conversion:
the anonymous function invocations were not flagged as using named
arguments. Therefore, the magic behind did not happened, which caused
the arguments not being stacked, and thus a weired runtime error.

(Also some space cleaning)

Issue #335, opened 2015-12-27
Named parameters generate error exception when used with returned function


k33g:
Named parameters generate error exception when used with returned function:

With this sample:
```golo
function yo = |message| {
  return |before, after| {
    println(before + ":" + message + ":" + after)
  }
}

function main = |args| {
  yo("bob")("Hello", "!") #Hello:bob:!
  yo("sam")(before="Hello", after="!")
}
```

`yo("sam")(before="Hello", after="!")` throw an IllegalAccessException:

    Exception in thread "main" java.lang.IllegalAccessException: no such method: named_params.main(String[])void/invokeStatic
            at java.lang.invoke.MemberName.makeAccessException(MemberName.java:872)
            at java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:993)
            at java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:1379)
            at java.lang.invoke.MethodHandles$Lookup.findStatic(MethodHandles.java:776)
            at org.eclipse.golo.cli.command.spi.CliCommand.callRun(CliCommand.java:39)
            at org.eclipse.golo.cli.command.GoloGoloCommand.execute(GoloGoloCommand.java:54)
            at org.eclipse.golo.cli.Main.main(Main.java:64)
    Caused by: java.lang.VerifyError: Operand stack underflow
    Exception Details:
      Location:
        named_params.main([Ljava/lang/String;)V @30: invokedynamic
      Reason:
        Attempt to pop empty stack.
      Current Frame:
        bci: @30
        flags: { }
        locals: { '[Ljava/lang/String;' }
        stack: { 'gololang/FunctionReference' }
      Bytecode:
        0x0000000: 1231 ba00 3900 00c0 0027 123b 123d ba00
        0x0000010: 4500 0057 1247 ba00 3900 00c0 0027 ba00
        0x0000020: 4500 0057 01b1
    
            at java.lang.invoke.MethodHandleNatives.resolve(Native Method)
            at java.lang.invoke.MemberName$Factory.resolve(MemberName.java:965)
            at java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:990)
            ... 5 more

Strangely, if I reverse the 2 lines:
```golo
function main = |args| {
  yo("sam")(before="Hello", after="!")
  yo("bob")("Hello", "!") #Hello:bob:!
}
```

I get a different exception:

[...]

PR #352, opened 2016-01-20
Compile Java sources with the -parameters flag

jponge:
Fix for #351

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,8 +48,8 @@ sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
 tasks.withType(JavaCompile) {
-  options.compilerArgs += ["-Xlint:unchecked", "-Xlint:deprecation"]
   options.encoding = "UTF-8"
+  options.compilerArgs += ["-Xlint:unchecked", "-Xlint:deprecation", "-parameters"]
 }
 
 // .................................................................................................................. //


jponge:
Surprisingly it fails the build :tada: 

In `namedarguments-anoncall.golo`, it seems like `invoke` calls on function references want `this` to be taken as a named parameter.

Any idea @yloiseau?

danielpetisme:
If you look in `FunctionCallSupport.reorderArguments` a mismatch appears.
* `Method`: `FunctionReference.invoke`
* `parameters`: `args`
* `arguments`: `this, before, after`

The `reorderArguments method` has a sanity check to be sure the argument names match parameter names. 
Obviously, this check doesn't find the argument `this` in the parameter names list {args}. You will get the same error with the `before` argument.

The weird thing is why it resolve to the `invoke` method and not the `$$_closure_XX` method...

yloiseau:
This change only imply changes in the way _Java_ methods are called. I think that is why only this `invoke` fails, since we are calling the `FunctionReference::invoke` method. However, this method is a generic varargs one, with `Object... args` as signature. These tests were added in the fix of #335 (PR #346), whose only significant change was to check argument names in `ParseTreeToGoloIrVisitor::visit(ASTAnonymousFunctionInvocation node, Object data)`.

This errors only shows that the test is buggy (shame on me). I only checked that named arguments are possible (the issue raised by @k33g showed that it was not even valid), but not that they are properly reordered. Indeed, if you try to switch named arguments (say `yo("sam"): invoke(after="!", before="Hello")`) the test fails (even in master branch) so the bug is not actually fixed in the `invoke` case (others seems to be OK). Since the `invoke` is generic, I'm not even sure it could be possible. And now that the names are checked due to the `-paramerters` flag, the issue becomes visible.

yloiseau:
I think we should simply discard the validity of calling `invoke` with named arguments, and remove the test. After all, `invoke` is a generic way to dynamically call _any_ function reference. 

If we want to deal with named arguments here, the runtime has to check the method handle of the receiving reference, instead of the `invoke` one, and then do a special case in `MethodFinder::toMethodHandle`.

danielpetisme:
I tried the following test :
`"JavaScript": substring(endIndex = 4, beginIndex = 0)`
it doesn't work, the reordering is not applied (`endIndex = 0, beginIndex = 4`) works. The Java method resolution relies only on the method signature and doesn't reorder args.

The aparently harmless `-parameters` flag raise a question. Should named argument work with Java or not?

yloiseau:
(@danielpetisme did you delete your previous comment?)
Damn... I thought it would work for fixed arity methods, but indeed, since we probably bypass the java mechanism in the invoke dynamic, it should not be surprising. Time for yet another refactoring of `MethodInvocationSupport` and `RegularMethodFinder` :angry: 

I think we should at least try to make it work for java fixed-arity functions, at least for consistency.

danielpetisme:
@yloiseau yes I deleted it because my test was not meaningfull.
I tried to use named args with a JDK class and since the JDK is not compiled with the `-parameters` flag it just can't work.

I made a test with a the Java `Predefined` class and it works `range(to = 'd', from = 'a')`.

I think we should raise a warning or exception when the dev uses named args with Java classes not compiled with `-parameters`.

yloiseau:
Ok. In that case, my remark is not meaningful either. Your tests made a point however, by highlighting the issue. Moreover, this issue will also raise when using 3rd party libs not compiled with `-parameters`. I had not envisioned this point. Too bad for consistency. 

:+1: for the warning.

yloiseau:
Does any other JVM language have named arguments? And if so, how do they deal with the native java aspect?

jponge:
Yeah, the JDK itself is not compiled with the flag, it's... bad https://julien.ponge.org/blog/java8-parameter-names-and-jdk-dogfooding/

danielpetisme:
to sum up,
1- the following snippet from `namedarguments-anoncall.golo` can't work
````
  assertEquals(yo("sam"): invoke(before="Hello", after="!"), expected)
  assertEquals(
    ^say: bindAt("message", "sam"): invoke(before="Hello", after="!"),
    expected)

  assertEquals(^add: invoke(left="a", right="b"), "ab")
  assertEquals((|left, right| -> left + right): invoke(left="a", right="b"), "ab")
````
because the true signature of `invoke` is `invoke(Object... args)` (no mention of `before`, `after`, `left` or `right`

So to fix the issue, we can't simply remove this tests

2- we have a bug when `invoke` is used in the proper way `invoke(args= ["Hello", "!"])`, it fails because it has a argument `this` not defined in the method formal parameters list.
We should create another issue to track it

3- Create an issue to discuss how manage properly named args with not incompatible Java classes (ie. not compiled with the `-parameters` flag.

PR #354, opened 2016-01-26
Fix a bug in Result::either

yloiseau:
The mapping function was not used, the recover one was used in both cases
(note to future me: don't use 42 in every test case)


jponge:
Looks good to me!

PR #357, opened 2016-02-02
Functions and operators utilities

yloiseau:
Adds a module containing useful functions to create
or change functions, to use in HOF for example.

Mainly operators as curried functions, composition functions, arity changing functions.

Adds a `toArray` and a `destruct` method to tuples, arrays, unions and structs to ease
polymorphism:

Adds tooling method to `FunctionReference`:
an `arity` function, and `invokeOrBind`: a function that apply arguments, binding them or
invoking the function according to the number.

### DIFF
--- a/src/main/golo/functions.golo
+++ b/src/main/golo/functions.golo
@@ -1,0 +1,1012 @@
+----
+This module contains several functions and higher order functions or
+decorators, mostly useful with HOF like `map`, `reduce` or in generators.
+
+Indeed, using HOF often require some classical functions like constants one,
+identity, or classical operators.
+
+Most of the times, these functions needs to be unary ones, i.e.
+have only one parameter. This module contains, among others, functions helping in the arity
+conversion of such functions:
+
+* [`curry`](#curry_f) that makes a function automatically partializable,
+* [`uncurry`](#uncurry_f) that is the reverse of `curry`,
+* [`unary`](#unary_f) and [`spreader`](#spreader_f) that convert a polyadic function into a unary
+  function taking an array or a tuple of values instead of several
+  values,
+* [`varargs`](#varargs_f) which is the reverse of `unary`
+
+Most of the binary operator functions defined here are curried. To ease
+partialization, the non commutative operators have a reversed version, with
+arguments swapped.
+For example, using the reverse division operator [`rdiv`](#rdiv_x_y) with
+[`map`](StandardAugmentations.html#java.util.List.map_this_func)
+
+    [2, 4, 8]: map(rdiv(2)) == [1, 2, 4]
+
+The identity function [`id`](#id_x), the constant function
+[`const`](#const_val), together with composition (`FunctionReference::andThen`,
+[`pipe`](#pipe_funcs), [`compose`](#compose_funcs)) and [`io`](#io_block) can be
+used to create monad-like processing chains.
+----
+module gololang.Functions
+
+import java.lang.invoke
+import java.util
+
+### Operators ##########################################################
+
+#== Arithmetic =========================================================
+----
+Polymorphic addition and concatenation.
+
+Adds `value` to `element`.
+The behavior depends on the type of `element`:
+
+- if a number or a string, equivalent to using `+`;
+- if a collection, equivalent to invoking `add` and returning the collection;
+- if an `Appendable` or a `StringBuilder`, equivalent to invoking `append` and
+  returning the collection.
+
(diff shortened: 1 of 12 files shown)


jponge:
This looks all very very good to me, great job Yannick! :+1:

yloiseau:
Thanks :blush: 
Comments are welcome, either on implementation details or on the general design/api/functions completeness

src/main/golo/functions.golo
```diff
@@ -0,0 +1,871 @@
...
+    list[1, 2, 3]: map(add(3)) == list[4, 5, 6]
+
+    add(42): andThen(add(1337)): andThen(add(69))(list[]) == list[42, 1337, 69]
+
+    add("a", StringBuilder()): toString() == "a"
+
+- *param* `value`: the value to add
+- *param* `element`: the element to add the value to
+----
+@curry
```
Artpej: According `curry` is pure and the only parameter is the decorated function ( i.e. `add` ) is constant, I think you can bang this decoration :wink:

Artpej: And the other ones too :smile:

yloiseau:
Some of the operators can fail (e.g. `div` or the destructuring `getitem` or `getter`) throwing exceptions. Should we make then to return `null` or `Optional` instead to allow easier composing or HOF use.
Users can still wrap them explicitly whith some decorators from `gololang.Errors`, but if the wrapping happens most of the time, it could be convenient to provide already wrapped functions. For instance, Rust `nth` (equivalent to `get`) returns an option type directly.
What do you think?

jponge:
Would a systematic wrapping make the code more complicated? If not then it would be in line with the goal of being functional.

yloiseau:
Well, it's just a matter of using the `nullify` or `option` decorators in `gololang.Errors`.
So for instance, the `getter` function, implemented as:
```golo
function getter = |i| -> |col| -> col: get(i)
```

would became 

```golo
function getter = |i| -> nullify(|col| -> col: get(i))
```

used as:

```golo
let g = getter(1)

g(list[1, 2, 3]) # 2
g(list[1]) # null

let k = getter("foo")
k(map[["foo", 42]]) # 42
k(map[]) # null
```

If the `getter` function is not wrapped, we can have the same result using:

```golo
let g = nullify(getter(1))
let k = nullify(getter("foo"))
```

On the other hand, not wrapping it let the final choice of the behavior to the user, e.g.:

```golo
let g = nullify(getter(1))
g(list[1, 2, 3]) # 2
g(list[1]) # null

let k = option(getter("foo"))
k(map[["foo", 42]]) # Optional[42]
k(map[]) # Optional.empty
```

jponge:
I'd vote to give the user the choice of what to do then.

danielpetisme:
:+1:

yloiseau:
No more comments?

jponge:
It's all good to me!

yloiseau:
Just a question about naming: the `inc` and `dec` (for increment and decrement) functions are sometimes named `pred` and `succ` (predecessor and successor). I think it could be more convenient, since `inc` and `dec` evokes side-effects,  while these functions have not. On the other hand, most developers may be more familiar with increment and decrement. 
What do you think?

k33g:
to my mind, `inc` and `dec` are more explicit

jponge:
`inc` and `dec` have mutation semantics, while `pred` and `succ` have immutable semantics, so I agree with Yannick here.

yloiseau:
To have some references, OCaml and F# have `incr` and `decr`, that take a "box" and have side-effects, and `pred` and `succ` that take an integer and have no effect. Haskell only has `succ` and `pred`, while Clojure uses  `inc` and `dec`. CLisp has `incf` and `decf` with side-effects, and `1+` without.

jponge:
What's the status of this PR Yannick?

yloiseau:
I just need to rename `inc` and `dec` as `pred` `succ` if we agree on that.

yloiseau:
I also want to refactor the memoizing decorator using bang and move it here but that can be done in another PR

jponge:
> I just need to rename inc and dec as pred succ if we agree on that.

I do :-)

jponge:
> I also want to refactor the memoizing decorator using bang and move it here but that can be done in another PR

As you feel like doing. There will be some days of lag between M2 and the formal 3.1.0 release, so we’ll be able to insert a PR or 2 if they don’t shake up things.

PR #361, opened 2016-02-12
Compile to Jar archives

jponge:
The golo compile used to work with an optional --output parameter to
specify a target directory.

With this change, if the target ends with '.jar', then the compilation
results are assembled to a Jar archive instead.

### DIFF
--- a/doc/basics.asciidoc
+++ b/doc/basics.asciidoc
@@ -147,6 +147,15 @@ classes/
 $
 ----
 
+It is also possible to output to a Jar archive:
+
+[source]
+----
+golo compile --output hello.jar samples/*.golo
+----
+
+This would take all `.golo` files from the `sample` folder, and assemble the resulting JVM class files in `hello.jar`.
+
 === Running compiled Golo code
 
 Golo provides a `golo` command for running compiled Golo code:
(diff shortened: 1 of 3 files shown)


jponge:
I am proposing this feature, what do the others think?

yloiseau:
Nice! :+1:

danielpetisme:
IMHO, produce a jar with the `compile` command is misleading. Why not add a
`package` command. The command could produce a simple jar or a fat one or
(because we are so hype ;)) a Docker image.

On ven., févr. 12, 2016 at 6:56 PM, Yannick Loiseau < notifications@github.com [notifications@github.com] > wrote:
Nice!:+1: [https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png]

—
Reply to this email directly or view it on GitHub
[https://github.com/eclipse/golo-lang/pull/361#issuecomment-183429797] .[https://github.com/notifications/beacon/AAkOnJwO2ntWUs9WPdsIkxK0AiHISpzUks5pjhRFgaJpZM4HZOAV.gif]

jponge:
Producing other kinds of artefacts like Docker images is likely out of the scope :-)

I personally think it is elegant to do things like this: the output of a compilation can be a directory, or it will be a Jar if the target ends with ‘.jar’.

We could add a kind of ‘package' or ‘jar’ task, but in the end of the day there already exists a tool for that: the good old `jar` command :-)
The goal here is really to make things easy.

jponge:
Merging!

PR #365, opened 2016-02-18
Add the ability to use either with Java 8

k33g:
It could be interesting to use golo as library for Java:

```java
import gololang.error.Result;
import java.util.function.Function;
...

Function<String, Result<Integer, RuntimeException>> toInt = (String val) -> {
  try {
    return Result.ok(Integer.parseInt(val));
  } catch (Exception e) {
    return Result.fail("Huston? We've got a problem!");
  }
};

Integer result = toInt.apply("fourty-two").either(value -> {
  System.out.println("Success: " + value);
  return value;
}, err -> {
  System.out.println(err.getMessage());
  return 42;
});
```

### DIFF
--- a/src/main/java/gololang/error/Result.java
+++ b/src/main/java/gololang/error/Result.java
@@ -9,16 +9,12 @@
 
 package gololang.error;
 
-import java.util.NoSuchElementException;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.List;
-import java.util.Iterator;
-import java.util.Collections;
+import gololang.FunctionReference;
+import gololang.Tuple;
+
+import java.util.*;
 import java.util.function.Function;
 import java.util.function.Predicate;
-import gololang.Tuple;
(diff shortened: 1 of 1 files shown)


jponge:
What's the link between your Java 8 code sample and using Golo as a library?

I understand what you have in your commit, but I can't see your point in the discussion :wink: :wine_glass:

k33g:
Java 8 brings us Optional, but not `Result`. I found it interesting to use
the `Result` of Golo, and especially the `either` method directly with Java.
But, to my mind the version of `either` with `FunctionReference` is hard to
use
And it's more "human" readable if we can pass `Function` as parameters of
`either` (the code becomes beautiful as JavaScript :p)

It's also an other way to demonstrate the interoperability between Golo and
Java, and by the way, it shows that Golo has some assets that not exist in
Java.

... Even if `Result` is written in Java

PR #366, opened 2016-02-23
FIX #249: use imported modules in fun

### LINKED ISSUE #249: Functions in imported module must be fully qualified to get a reference
To get a reference to a function defined in an imported module using the `^` operator, it *must* be fully qualified, contrary to its usage.

For instance, given the module:
```golo
module foo

function bar = -> "bar in foo"
```
and a module using it:
```golo
module baz
import foo

function main = |args| {
  println(bar())

  let fqn_ref = ^foo::bar
  println(fqn_ref())

  let ref = ^bar
  println(ref())
}
```

The direct call works as expected, as the fully qualified reference. But the last assignment yields a 
`NoSuchMethodException: bar in class baz`, which seems counter intuitive since it is “visible in the local scope” (or act as if).

yloiseau:
- small refactor of `Predefined::fun`:
    just use existing methods from `Extractors` and `TypeMatching`;
- add a `varargs` parameter to `fun` to specify if a variable arity
  method is required or not;
- search in modules imported by the one we are looking in if no function
  is found (use the first function found).

### DIFF
--- a/src/test/java/gololang/PredefinedTest.java
+++ b/src/test/java/gololang/PredefinedTest.java
@@ -145,13 +145,13 @@ public class PredefinedTest {
 
   @Test
   public void test_fun() throws Throwable {
-    FunctionReference hello = (FunctionReference) Predefined.fun("hello", MyCallable.class, 0);
+    FunctionReference hello = Predefined.fun("hello", MyCallable.class, 0);
     assertThat((String) hello.handle().invoke(), is("Hello!"));
   }
 
   @Test
   public void test_fun_no_arity() throws Throwable {
-    FunctionReference hello = (FunctionReference) Predefined.fun("hello", MyCallable.class);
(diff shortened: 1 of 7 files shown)


src/test/java/gololang/PredefinedTest.java
```diff
@@ -162,24 +162,24 @@ public void test_fun_fail() throws Throwable {
 
   @Test(expectedExceptions = AmbiguousFunctionReferenceException.class)
   public void test_fun_ambiguous() throws Throwable {
-    MethodHandle overloaded = (MethodHandle) Predefined.fun("overloaded", MyCallable.class);
```
jponge: Ooops... fortunately it failed before the cast was even attempted :smile:

jponge:
Very good @yloiseau I'm merging it.

Issue #367, opened 2016-02-23
Update golo project skeletons


danielpetisme:
The `golo new` command generates a Gradle or Maven project based on a skeleton. These skels are out of date:
* use the old `org.golo-lang` groupId
* use maven central as default repository for plugins instead of jcenter
* use out-of date golo language and plugin versions (before the move to Eclipse)

jponge:
:+1:

jponge:
Note: Maven Central is still fine, the artifacts are available there.

yloiseau:
:+1:

yloiseau:
see also https://github.com/golo-lang/gradle-golo-plugin/issues/6

jponge:
@danielpetisme are you willing to fix it for milestone 2?

danielpetisme:
Not sure to have time, I can't commit on a date.

jponge:
Hum... I just checked and it seems all current:

* https://github.com/eclipse/golo-lang/blob/master/src/main/resources/new-project/maven/pom.xml#L38
* https://github.com/eclipse/golo-lang/blob/master/src/main/resources/new-project/gradle/build.gradle

danielpetisme:
I checked with the latest release.
The groupIds are ok but there is a `${project.version}` not filtered
https://github.com/eclipse/golo-lang/blob/master/src/main/resources/new-project/maven/pom.xml#L14
This is what I have when I create a Golo Maven project

```
  <groupId>org.company</groupId>
  <artifactId>Golo</artifactId>
  <version>0.0.1-SNAPSHOT</version>
  <packaging>jar</packaging>

  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <golo.version>${project.version}</golo.version>
  </properties>

  <dependencies>
    <dependency>
      <groupId>org.eclipse.golo</groupId>
      <artifactId>golo</artifactId>
      <version>${golo.version}</version>
    </dependency>
  </dependencies>
```

PR #370, opened 2016-03-10
Fix for #367

jponge:
Process the Gradle and Maven templates to have the Golo version edited.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -111,8 +111,9 @@ test {
 
 processResources {
   filter(org.apache.tools.ant.filters.ReplaceTokens, tokens: [
-    version: version,
-    timestamp: versioning.info.full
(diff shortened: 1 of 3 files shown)


jponge:
Ping @danielpetisme re: #367

danielpetisme:
How can we test automatically such thing?
Otherwise I'll do a manual test tonight.

jponge:
> How can we test automatically such thing?
> Otherwise I'll do a manual test tonight.

`cat build/resources/…` :-)

Otherwise automated testing for that is about opening the resource, and playing with regexps.

danielpetisme:
Tested, it works.

Regarding automation, is tricky to grep from a Java test the `build.gradle` version...

PR #375, opened 2016-03-31
Extended literal notation for floating numbers

### LINKED ISSUE #374: Native BigDecimal and BigInteger support
Make support of other Java `Number` subclasses (namely `BigInteger` and `BigDecimal`) native. This would imply:
* a literal value notation, e.g. `3.1415927_B` for `BigDecimal`;
* adapt arithmetic operators to use the Java method: `1.5_B + 0.5_B` would be converted into `BigDecimal.valueOf(1.5).add(BigDecimal.valueOf(0.5)`, with an *implicit* conversion into `BigInteger` or `BigDecimal` when operators are used on mixed types;
* *implicit* conversion when comparing numbers: indeed, `BigDecimal` (resp. `BigInteger`) implements `Comparable<BigDecimal>`, and thus can't be compared to other numbers. For instance, currently:
```golo
BigInteger.ONE() > BigInteger.ZERO()   # true
BigInteger.ONE() > BigDecimal.ZERO()   # does NOT work
BigInteger.ONE() > 0                   # does NOT work
```

Maybe:
* a way to define global default `MathContext` to be used;
* use `BigDecimal` as default non-integer type (as Groovy do), with a `_D` suffix to force `Double`

yloiseau:
Fix #374 by:
* allowing `_` in floating numbers and negative exponent in scientific notation;
* adding literal notation for `BigInteger` and `BigDecimal` using the `_B` suffix;
* extending operators to deal with `BigInteger` and `BigDecimal` values.

### DIFF
--- a/doc/basics.asciidoc
+++ b/doc/basics.asciidoc
@@ -272,6 +272,10 @@ API. We give them along with examples in <<data-literals,the data literals table
 
 |`java.lang.Float` | `1.234_F`, `-1.234_F`, `1.234e9_F`, ...
 
+|`java.math.BigInteger` | `1_B`, `-42_B`, `1_234_B`, ...
+
+|`java.math.BigDecimal` | `1.0_B`, `-1_234.56_B`, `1.234e-4_B`, ...
+
 |`java.lang.Class` | `String.class`, `java.lang.String.class`, `gololang.Predef.module`, ...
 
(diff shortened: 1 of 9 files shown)


jponge:
Everything looks good to me, the only missing piece now is a documentation update :smiley:

jponge:
All good!

PR #377, opened 2016-04-19
Check command

### LINKED ISSUE #324: New check command line command
Add a new `golo check` command, similar to `golo compile`, that parse input files and run the compile time IR verifications without generating the `.class` files. Useful for instance to integrate with editor plugins such as Vim syntastic.

yloiseau:
Adds a `check` command to the `golo` binary. This command does the same
tasks as the `compile` command, namely parse input files and run the
compile time IR verifications, but does not produce result files.

This command can be useful to integrate golo with editor plugins such as
Vim Syntastic.

FIX #324


jponge:
👍

Looks good to me!

PR #378, opened 2016-04-20
FIX: bug when checking several files

yloiseau:
When processing several files, and without the `--exit` option, only the
first error was displayed, for every file with an error, because the
exception builder was not reset between files.

Also adds a verbose option to display the checked file name.


yloiseau:
oops

jponge:
+1 :-) 
> 
> 
> 
>

PR #381, opened 2016-05-08
Run Golo application from a shebanged script

Artpej:
ref: [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix))

This feature aim to run golo application as simple runnable script.

```
golo shebang /path/to/script.golo arg1 arg2
```
this command do:
 - compile every golo files in the `/path/to/` directory recursively
 - add every `jar` files in the `/path/to/` directory recursively to the current classpath
 - run `script.golo` main funtion with `/path/to/script.golo`, `args1` and `arg2` as parameters

so we can now shebang this script file (script.golo):
```golo
#!/path/to/golo  shebang
module script

function main = |args| {
  foreach arg in args {
    println(arg)
  }
}

```
then in a shell `./script.golo arg1 arg2` will print
```
./script.golo
arg1
arg2
```

### DIFF
--- a/doc/basics.adoc
+++ b/doc/basics.adoc
@@ -198,7 +198,7 @@ Naturally the main goal is to use this command to make the script self-executabl
 
 [source,golo]
 ----
-#! /path/to/golo shebang
+#!/path/to/golo shebang
 module hello
 
 function main = |args| {
@@ -221,7 +221,7 @@ a golo script can be hasbanged with `env`:
 
 [source,golo]
 ----
-#! /usr/bin/env golosh
+#!/usr/bin/env golosh
 module hello
 
 function main = |args| {
--- a/doc/basics.adoc
+++ b/doc/basics.adoc
@@ -171,6 +171,85 @@ $
 
 Simple, isn't it?
 
+=== Running Golo script
+
+Golo provides a `shebang` command for running a Golo file as a simple script.
+
+[source,golo]
+----
+module hello
+
+function main = |args| {
+  require(args: length() > 1, "You should set at least one argument!")
+  println("Hello " + args: get(1) + " from '" + args: get(0) + "'!")
+}
+----
+
+the script above can be executed with:
+
+[source]
+----
+$ golo shebang hello.golo World
+Hello World from 'hello.golo'!
+$
+----
+
+Naturally the main goal is to use this command to make the script self-executable:
+
+[source,golo]
+----
+#! /path/to/golo shebang
+module hello
+
+function main = |args| {
+  require(args: length() > 1, "You should set at least one argument!")
+  println("Hello " + args: get(1) + " from '" + args: get(0) + "'!")
+}
+----
+
+Now, we can run the script directly:
+
+----
+$ chmod +x hello.golo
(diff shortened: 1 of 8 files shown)


Artpej:
It is just a proposal for now, WDYT?

- [x] first implementation (POC)
- [ ] test
- [ ] documentation
- [ ] refactor the Golo classloader to cache already compiled golo files?

yloiseau:
:+1:

jponge:
Cool stuff!

yloiseau:
Regarding cached compiled files, I see two options:
- save the `.class` in the same directory as the original `.golo`, like python does for instance;
- save the `.class` in a "global" user directory (e.g. `$HOME/.cache/golo` under [Linux XDG](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)).

The global cache has the advantage that is can simply be added to `$CLASSPATH` once for all. 
The `compiler.compileTo` method can be used, just as is the `compile` command.

However, two issues here: 
- class finding;
- time checking.

While time checking can be quite easy, finding the `.class` to compare to is less trivial.
Indeed, in Golo (unlike python e.g), the output file name is derived from the golo _module_ name, and not the _source file_ name. Thus, in order to know if the given file has already been compiled, it needs to be parsed anyway, which reduce the purpose of caching.

jponge:
Any update on this PR?

Artpej:
Yes, I'll complet this one soon with doc and some tests. 
Add cache on generated bytecode will be a part of another PR.

jponge:
Makes sense to split into 2 PRs 👍

jponge:
Ping?

Artpej:
All right for the review.

doc/basics.adoc
```diff
@@ -171,6 +171,85 @@ $
...
+$ golo shebang hello.golo World
+Hello World from 'hello.golo'!
+$
+----
+
+Naturally the main goal is to use this command to make the script self-executable:
+
+[source,golo]
+----
+#! /path/to/golo shebang
```
jponge: Shouldn't we go without a space, like `#!/usr/bin/env golosh`?

doc/basics.adoc
```diff
@@ -171,6 +171,85 @@ $
...
+Hello World from 'hello.golo'!
+$
+----
+
+NOTE: Golo also provides `golosh` script that is a shortcut for the `golo shebang` command, thus
+a golo script can be hasbanged with `env`:
+
+[source,golo]
+----
+#! /usr/bin/env golosh
```
jponge: Same question here

jponge:
Minor comments, otherwise looks good to me!

doc/basics.adoc
Artpej: It works for me with a space... it seemed to me more readable like this, but you'r right I've never seen a script whit a space here

jponge:
Ah @Artpej that did not pass the IP checks.

Artpej:
@jponge since few days ip validator is not my friend... (my bad, to many computers, I should sync my dot files :wink:)

jponge:
👍

PR #384, opened 2016-05-28
Add unicode escaped chars in `strings` and `chars` literals.

Artpej:
Add unicode escaped chars in `strings` and `chars` literals.


jponge:
Go for it!

PR #399, opened 2016-08-12
Fix conflict between property and union attribute resolution

### LINKED ISSUE #398: Getting union type attribute returns "true"
When an union type attribute has the same name that his owner (except uppercase for the first char), then getting this attribute is not possible (always returns true):

```golo
union U = {
  Abc = { abc }
}

function main = |args| {
  let x = U.Abc("a")
  println(x: abc()) # will print "true" ?!
  # because: x: abc() == x: isAbc()
}
```

In this case `x: abc() ` behave like `x: isAbc()`. I didn't found anything in the documentation about this  behaviour.

:smile_cat:  So, bug or feature?

Artpej:
This fixes #398


jponge:
👍

Issue #407, opened 2016-10-11
Compiler hangs when declaring variable without assigning a value


vvdleun:
When declaring a variable, but not assigning a value at the same time, the compiler seems to hang.

The following script hangs when compiling:

```
module Test
 
function main = |args| {
       var t
}
```

I have seen this behavior in Golo v3.2.0-M4 and I believe this happens in 3.1.0 as well.

Please let me know if I can be of any further help.

yloiseau:
Ok, my bad, I introduced the regression in https://github.com/eclipse/golo-lang/commit/2f1dc71787506edd45fa1d1242684cfbbb03b940 when adding destructuring (thanks `git bisect`)

PR #408, opened 2016-10-11
FIX: #407 Compiler hangs on unassigned variable

yloiseau:
Fix a bug introduced in https://github.com/eclipse/golo-lang/commit/2f1dc71787506edd45fa1d1242684cfbbb03b940 and identified by @vvdleun in #407 
When declaring a name with `let` or `var` but affecting no value, the parser hangs.

The issue was a lookahead to high for this expression, added for destructuring.

### DIFF
--- a/src/main/jjtree/org/eclipse/golo/compiler/parser/Golo.jjt
+++ b/src/main/jjtree/org/eclipse/golo/compiler/parser/Golo.jjt
@@ -1006,7 +1006,7 @@ void Statement() #void:{}
   |
   LOOKAHEAD(3) DestructuringAssignment()
   |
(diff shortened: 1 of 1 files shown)


yloiseau:
maybe we should add test for the original bug, but I have no idea how to test this, since the consequence is to hang...

rlespinasse:
Maybe a timeout test

jponge:
That one is hard to test since the fix removes the hang 😉 

One option we could have for all `CompilationTest` is to introduce a _timeout_, and reproduce that specific bug in the tests suite, hence if we ever have a regression again the tests suite will fail instead of getting stuck.

PR #409, opened 2016-10-12
Allow to get primitive types with .class operator

Artpej:
Useful to deal with primitive Java type

for exemple : `newTypedArray(byte.class, 1024)`


src/main/java/org/eclipse/golo/runtime/ClassReferenceSupport.java
```diff
@@ -55,7 +55,17 @@ public static CallSite bootstrap(MethodHandles.Lookup caller, String name, Metho
     try {
       return Class.forName(name, true, classLoader);
     } catch (ClassNotFoundException e) {
-      return null;
+      switch(name) {
```
yloiseau: An other elegant approach is to use a `Map<String, Class>` and thus only `return PRIMITIVES.get(name)` instead of the `switch`, but I suspect it would be less efficient.

jponge: It's bootstrap-time, so a map lookup won't be a performance issue.

jponge: Another option is to avoid a map allocation (even if it's small) and refactor that `switch` to a `private` method. That's perhaps even more elegant.

Artpej:
you're right, this just a quick patch I've done for a side project requirement, sorry I forgot mention that this work is still in progress :wink:.

I will do some refacto soon ;)

src/main/java/org/eclipse/golo/runtime/ClassReferenceSupport.java
Artpej: @yloiseau use `Map<String, Class>` cache seems to me not a good idea according `Class` instances are classloader dependent and some project use multiple classloaders (even if there really is no case primitive types to be in two different classloaders)
@jponge yes, I had in mind to extract a method and check first if it is a primitive type (not in the catch clause)

Artpej:
Hi Travis , are you alive?:wink:

jponge:
Looks good, I think we only lack some documentation update now.

jponge:
Looks good to me!

PR #414, opened 2016-11-10
FIX #248: local type must be fully qualified

### LINKED ISSUE #248: Local struct must be fully qualified to access its class
To access the class object of a locally defined struct, it _must_ be fully qualified, contrary to other classes.
For instance, given the module:

``` golo
module foo

struct Foo = {a}

function main = |args| {
  println(Tuple.class)
  println(String.class)
  println(foo.types.Foo.class)
}
```

outputs the expected:

```
class gololang.Tuple
class java.lang.String
class foo.types.Foo
```

but if we use instead 

``` golo
  println(Foo.class)
```

we got a `BootstrapMethodError`. An explicit `import foo.types` is required, which seems counter intuitive.

yloiseau:
This PR fixes the #248 bug.

When defining new types (`struct` or `union`), the classes are compiled in a `[moduleName].types` sub-package, that is not directly accessible from the defining module. As a result, the class can't be accessed directly without fully qualifying it or importing the `types` sub-package.

For instance, the following code fails

```golo
module foo

struct Foo = {a}

function main = |args| {
  println(Foo.class)
}
```

with a `BootstrapMethodError` since `foo.types` is not looked up.

The proposed fix is quite simple: it adds an `import [currentModule].types` as the last `import` statement to every module.


yloiseau:
I added the import as the last one, but I was thinking that since the first matching import is used and we prioritize local declarations, it would be better to insert it as the first import...

jponge:
Yep ✅

Le jeu. 10 nov. 2016 à 20:54, Yannick Loiseau <notifications@github.com> a
écrit :

PR #416, opened 2016-11-13
Change the golodoc html title

yloiseau:
When a lot of tabs are opened, the title was shortened as "Documentation
of..."
The module name is moved at the beginning such that it is displayed
even when shortened.


Artpej:
:+1: , add a favicon is somthing else that could be very helpful too :wink: 

We could use a "`data:image/x-icon;base64...`"  in the link source to keep the documentation avaiability offline.

just ping, if you want me to amend this PR with this.

yloiseau:
I put an hard coded String in the HTML template, but I wonder if it wouldn't be better to put a resource file and generate the base64 on the fly...

yloiseau:
Does anyone know how to customize the asciidoctor template to also bundle a icon in the golo guide as suggested by @Artpej ?

yloiseau:
Are we OK with this one ?

jponge:
Yes!

PR #417, opened 2016-11-16
Adds a `catcher` HOF in gololang.Errors

yloiseau:
This HOF create a function that can execute a block and apply a given
function if an exception occurs. It can be used as a decorator.


jponge:
Looks good to me!

Issue #383, opened 2016-05-27
Improved golo doc


yloiseau:
I think we could improve `golo doc` in several ways, by looking at what is done in Haskell and Rust (see for instance https://doc.rust-lang.org/std/option/ and https://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Maybe.html)

For instance:
- Haskell links the functions documentations to a html version of the source code;
- the snippets in the Rust documentation can be run directly in an online interpreter (try to hover the samples);
- the snippets in the Rust documentation are used as tests in the compile step (similar to Python doctest https://docs.python.org/3/library/doctest.html).

What do you think?

jponge:
I concur on all 3 ideas!

PR #418, opened 2016-11-17
Link golodoc and source file

yloiseau:
This implements the first point in #383: the golodoc now links to a html version of the source file, to the line where the documented element is.


jponge:
Excellent!

PR #420, opened 2016-12-01
Move the JavaCC/JJTree generated code to src/main/generated

jponge:
This makes it easier to locate these files from an IDE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ tags
 build/
 .gradle/
 /.nb-gradle/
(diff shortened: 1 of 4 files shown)


yloiseau:
Are the generated files still removed by the `clean` task?

jponge:
Good point, no, they aren't...

jponge:
@yloiseau I've added a new cleaning task.

yloiseau:
:+1:

PR #423, opened 2016-12-08
Changes to DynamicObject

yloiseau:
Adds a `kind` attribute that can be used to compare dynamic objects and
in the toString method.
Adds a toString method that display all fields.

### DIFF
--- a/doc/dynamic-objects.adoc
+++ b/doc/dynamic-objects.adoc
@@ -12,6 +12,13 @@ Creating a dynamic object is as simple as calling the `DynamicObject` function:
 let foo = DynamicObject()
 ----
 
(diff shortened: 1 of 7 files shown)


yloiseau:
I always forget the documentation :smile:

yloiseau:
Not ready yet, I found 2 bugs while writing the documentation! Always test in golo, not in java...
I have not updated the dynamic dispatch reserved methods so `sameKind` and `hasKind` set properties...
And since `toString` prints properties, if we have cyclic references we stack overflow... not sure how to fix this one (this is probably the case for struct and union too)

yloiseau:
I think I'm OK for final review

Issue #419, opened 2016-11-18
Default values for struct members


yloiseau:
It would be nice to be able to specify default values for `struct` members, more specifically when one of this member should be a container.

### Proposed syntax

```golo
struct Foo = {
    a,  # required value
    b=42, # default value
    c=list[]
}
```

### Implementation idea

Currently, a `struct` declaration generate two construction functions in the module:
```golo
function Foo = -> ...
function Foo = |a, b| -> ...
```

in the presence of default value, we could generate another one:
```golo
function Foo = |a| -> Foo(a, 42, list[])
```

this is just a compile time desugarization, which *should* be quite simple.

jponge:
Would be a great addition.

How about `enum`?

yloiseau:
`union` constructors are not of the same kind, to keep the namespace. For instance, given
```golo
module MyModule
union Foo = {
  Bar = {plic, ploc}
  Baz
}

union GoodPlace = {
  Restaurant
  Bar = {name}
}
```

we can't generate the `Bar` function. The constructor is a static function in the `Foo` and `GoodPlace` classes, to be used as:
```golo
let meeting = GoodPlace.Bar("Still Irish Bar")
```
As a side note, this requires the import of `MyModule.types` to use the `union`s of `MyModule` like that (whereas it's not require for `struct`s) or to use FQN:
```golo
let plop = MyModule.types.Foo.Bar("answer", 42)
```

Moreover, since `union` types are designed to be immutable, I'm not sure if a default value is relevant. Obviously, we could put a mutable object in the union value, so the immutability can't be fully guarantied. 

However, it *should* not be that hard to add the feature for `union`, and I'm sure some could find a interesting use case, so why not :smile:

yloiseau:
Experimenting with this. When we have

```golo
module Bar
struct Foo = { x, y }
```

2 constructor are generated (resp. Immutable)

```golo
function Foo = -> Bar.types.Foo()
function Foo = |x, y| -> Bar.types.Foo(x, y)
```

With defaults, we generate 

```golo
struct Foo = { x, y=42}
function Foo = -> Bar.types.Foo(null, 42)
function Foo = |x| -> Bar.types.Foo(x, 42)
function Foo = |x, y| -> Bar.types.Foo(x, y)
```

So far so good. However, I'd like to allow what I call "dependant default":

```golo
struct Foo = {x, y=2*x}
```

that would generates

```golo
[...]

Issue #189, opened 2014-10-02
Default values on function parameters


danielpetisme:
Wouldn't be nice to do 

``` golo
function split = |text, seperator=";"| {
...
}

require(split("foo;bar") is ["foor", "bar"], "Default separator value")
require(split("foo:bar", ":") is ["foor", "bar"], "Explicited defined separator value")

```

danielpetisme:
Question about this one.

Currently, the function resolution is based on the function name + number of args (+ types when applicable). When a parameter has a default value the argument is optional and the resolution is a bit harder.

````
//let's call it split_1
function split = |text, separator=";"| {
...
}

//let's call it split_2
function split = |text| -> split(text, "-")

// which one should be called ?
split("foo;bar") 
````

1- Explicit > Implicit: name match and nbArgs match => split("foo;bar")  invokes `split_2`
2- Partialization first, invokes the longer signature with default values ?

@yloiseau how is it handled in Python?

**Comment by jponge** [comment]
In that case `split("foo;bar")` must call `split_2`.

An implementation of default arguments shall detect function arity clashes in a module and report errors at compilation time IMHO.

**Comment by jponge** [comment]
@danielpetisme This is not supported in Python, a function gets redefined:

```python
def a(x, y=0):
    return x + y

def a(n):
    return n * 100

print a(1)
print a(1, 2)
```

Gives:

```
python foo.py 
100
Traceback (most recent call last):
  File "foo.py", line 8, in <module>
    print a(1, 2)
TypeError: a() takes exactly 1 argument (2 given)
```

danielpetisme:
So given:
`
function split = |text, separator = ";"| -> ...
function split = |text| -> ...

split("foo:bar")
`

this snippet should raise a kind of `Ambiguous resolution exception`

Right ?

jponge:
That would be sane :-)

danielpetisme:
Other point.
What's the best option.
Build a Parameter structure which holds the parameter name and the default value, when the bytecode is generated pass the default values in the BSM arguments (just as named args are working right now) => The default values injection is made at compilation time.

Or at compilation time generate the partialized function
````
function split = |text, separator| -> ...
//would produce the bytecode for
function split = |text, separator| -> ...
function split = |text| -> split(text, ";")
````

At last but not least, do we accept constants or variables/expression are allowed.
````
function split = |text, separator = System.lineSeperator()| ...
````

yloiseau:
[...]

PR #426, opened 2016-12-14
Adds default values for fields in struct and union

yloiseau:
This PR allows to define default values for fields in struct and union, as described in #419.

It defines a new IR element, `Member`, to encapsulate the name and default value of the field, as well as a new abstract node `TypeWithMembers` to define common behavior for struct and union values regarding their fields.

Since we have an AST object, it is now possible to document fields:

```golo
struct Foo = {
  ----
  the foo
  ----
  foo,

  ----
  the bar
  ----
  bar
}
```

These comments are displayed in the golodoc.

The main feature is implemented by creating alternative factory methods in the module declaring the struct or the union. See #419 for some details.
For instance, a struct defined as:

```golo
struct Foo = {a, b=42, c=f(a)}
```

will create several factories:

```golo
function Foo = |a, b, c| -> module.types.Foo(a, b, c)
function Foo = |a| -> module.types.Foo(a, 42, f(a))
function Foo = -> raise("Can't call the default constructor...")
```
while
```golo
struct Bar = {a, b=42}
```
will generate:
```golo
function Bar = -> module.types.Bar(null, 42)
function Bar = |a| -> module.types.Bar(a, 42)
function Bar = |a, b| -> module.types.Bar(a, b)
```


jponge:
I'm also asking reviews from @Artpej for his expertise, and @k33g for his unique ability to spot improbable bugs 😈

doc/structs.adoc
```diff
@@ -101,6 +101,7 @@ function main = |args| {
 }
 ----
 
+[[struct-curstom-factories]]
```
jponge: custom

doc/structs.adoc
```diff
@@ -307,6 +308,60 @@ Private struct members also have the following impact:
...
+When declaring a `struct`, default values can be given for members. These values will be used to initialize instances when no explicit values are given. For instance:
+
+[source,golo]
+----
+module Test
+struct Foo = {a, b=42}
+
+# (...)
+
+println(Foo(0))
```
jponge: BTW can we also use named arguments?

doc/structs.adoc
```diff
@@ -307,6 +308,60 @@ Private struct members also have the following impact:
...
+will print `struct FooDepend{a=15, b=27, c=42}`. Any valid golo expression using other non-default members can be used.
+
+NOTE: This is not an checked invariant, but only a default initialization. If `c` must always be `a + b`, just create an <<struct-augmentation,augmentation>> to compute the value.
+
+WARNING: When using _dependant default values_, the default constructor, which usually initialize non default members to `null` will raise an exception, even if the default expression could deal with `null` values. It is always possible to define a custom factory to implement the required behavior.
+
+WARNING: A _dependant default value_ can't depend on a member having a default value itself. To create complex default logic, you are encouraged to create your own <<struct-custom-factory,custom factory>>.
+
+
+[[strct-augmentation]]
```
jponge: Typo?

Artpej:
same as @k33g from me:

`java.lang.AssertionError at CompileAndRunTest.java:1909`

```
java.lang.AssertionError: 
Expected: <class java.lang.ClassNotFoundException>
     but: was <class java.lang.NoClassDefFoundError>
```

the same issue happen for me on the `master` branch (with a Windows host) :open_mouth:

Artpej:
whatever, this test issue is not related to this PR (in fact it's related to an old one of mine :confused: )

yloiseau:
Damned... @Artpej only on Windows?

Artpej:
@yloiseau I don't know ( @k33g ? ), but my linux computer didn't behaved like this and Travis too.

k33g:
@Artpej @yloiseau on OSX

Artpej:
@yloiseau I exepected the following code to be correct but It seems that only factories are improved, what about setters?

```golo
  let setter = MultiDepStruct(1, 4): a(1): b(2)
  assertThat(setter: a(), `is(1))
  assertThat(setter: b(), `is(2))
  assertThat(setter: c(), `is(3)) # wrong : c = 5
```

Antother thing: default factory `MultiDepStruct()` is no more avaiable : `Can't call the default constructor of a structure with dependant default value.`, maybe when a required avlue is null the computed one can be null too.

thought?

yloiseau:
@Artpej, indeed, this is the intended behavior. For your first point, see https://github.com/eclipse/golo-lang/pull/426/files#diff-12133d45818249dbff2cc02a2a96fbb4R357
For the second one, see https://github.com/eclipse/golo-lang/issues/419#issuecomment-261877284 and 
https://github.com/eclipse/golo-lang/pull/426/files#diff-12133d45818249dbff2cc02a2a96fbb4R359

We can indeed call the default constructor instead of throwing an exception, but I found it less consistent. Suppose you define a struct like:
```golo
function f = |v| -> v?: toString() orIfNull ""
struct Foo = {a, b=f(a)}
```
`Foo(42)` is `struct Foo{a=42, b="42"}`. But it would be legitimate to expect `Foo()` to be `struct Foo{a=null, b=""}`. However, since I can't tell compile time that `f` can be called on `null`, I can either don't call if and then `Foo()` is `Foo{a=null, b=null}` which in my POV can be misleading, or call it anyway, and get an more obscure exception if `f` does not deal with `null` correctly. I chose to throw early with a more explicit message.

doc/structs.adoc
yloiseau: Yes. I'll add a note in the doc and a case in the tests.

doc/unions.adoc
```diff
@@ -182,6 +182,27 @@ function dealWithTree = |elt| -> match {
...
 ----
 
+=== Default values
+
+Union values members can have default values, in the same way as <<struct-default-values,`struct`>>.
+For instance:
+
+[source,golo]
+----
+union List = {
```
k33g: My understanding is that I can write something like that:
```golo
union SomeThing = {
  Yum = { value="🍊" }
  Yuck = { value="🍋" }
}
```
and then I hope(expect?) to get:
```golo
Something.Yum("🍆") is union SomeThing.Yum{value=🍆}
Something.Yum() is union SomeThing.Yum{value=🍊}
```

but the call of `Something.Yum()`
generates error:
```
Exception in thread "main" java.lang.invoke.WrongMethodTypeException: cannot convert MethodHandle(Object)SomeThing to ()Object
	at java.lang.invoke.MethodHandle.asTypeUncached(MethodHandle.java:775)
	at java.lang.invoke.MethodHandle.asType(MethodHandle.java:761)
	at org.eclipse.golo.runtime.FunctionCallSupport.fallback(FunctionCallSupport.java:164)
	at main.main(main.golo:33)
	at org.eclipse.golo.cli.command.spi.CliCommand.callRun(CliCommand.java:40)
	at org.eclipse.golo.cli.command.ShebangCommand.execute(ShebangCommand.java:46)
	at org.eclipse.golo.cli.Main.main(Main.java:62)
```

I probably missed something

yloiseau: Or you found a bug. That's indeed the expected behavior. I'll have a look

yloiseau: Ah got it. It's a documentation bug :-)
When you call `Something.Yum()`, you call the java factory method, which is not redefined by this PR. As for the struct, factory golo functions are generated, so you have to call `Yum` (without prefixing)

k33g: ok, it works, but 😢 I thought it was nice because of readability (IMHO)

yloiseau:
@Artpej in my point of view, default values as defined here are just, that: default values for the attributes if none are given to the constructor, much like in java `private String name = "Toto"; `. They are not intended to be a constraint that the attribute must hold, which is more like the solution you propose, or more something like properties in python or C# (or ES6 IIRC). In one of your previous comment:
```golo
struct MultiDepStruct = {a, b, c=a+b}

let setter = MultiDepStruct(1, 4): a(1): b(2)
assertThat(setter: a(), `is(1))
assertThat(setter: b(), `is(2))
assertThat(setter: c(), `is(3)) # wrong : c = 5
```

If this is the intended behavior, that is, `c` is *always* `a + b`, it could be simply defined as:
```golo
struct MultiDepStruct = {a, b}
augment MultiDepStruct {
  function c = |this| -> this: a() + this: b()
}
```

or with cached value:
```golo
struct MultiDepStruct = {_a, _b, _c}
augment MultiDepStruct {
  function a = |this| -> this: _a()

  function a = |this, v| {
    this: _a(v)
    this: _c(v + this: b())
  }

  function b = |this| -> this: _b()

  function b = |this, v| {
    this: _b(v)
    this: _c(this: a() + v)
  }
  function c = |this| -> this: _c()
}
function MultiDepStruct = |a, b| -> MultiDepStruct(): a(a): b(b)
```
(`a` and `b` are private since we can't redefine native methods in augment).

Indeed, `c` must be read only, because we can't define a setter that dispatch changes to `a` and `b` in the general case. 
However, it could indeed be interesting to define syntactic sugar for such cases, but it's beyond the scope of this PR (imo).

yloiseau:
The bang issue raised by @Artpej is less trivial and require more discussion.

I think I'll split this PR in two, one with only the `Member` refactoring to keep the possibility to document members and attach meta-data, and another one for this default value feature, which is not as clear as I thought.

yloiseau:
Maybe this feature should use #189 and just define default parameters in the factory function.

PR #427, opened 2016-12-18
Add compiler option to set UTF-8 as default encoding

Artpej:
When I try to compile the Golo project on an OS that does use UTF-8 as default encoding some warnings are displayed beacause of some UTF-8 chars present in Java comments and javadoc.
This one fixes this issue.


jponge:
Oh yes!

PR #429, opened 2016-12-23
Refactor reordering of named parameters

yloiseau:
partial fix for issues raised in #351 and #352 comments.

- deal with the receiver in method calls: the receiver argument has no
name; moreover, for regular methods, the receiver parameter has no name
(but is present in the signature)

- display a warning if calling a function without parameter names with
named arguments. It can be disabled via properties.


yloiseau:
@jponge I think that with this one you can go on #352

jponge:
`BUILD SUCCESSFUL` when applying #352, awesome 👍 😃

PR #431, opened 2017-01-10
Adds an index file to the golodoc

yloiseau:
The index file lists all elements wherever they are defined (functions,
struct, unions, augmentations, and so on) with a direct link to the
documentation of the corresponding element.
The elements are grouped by their first letter, with a way to filter the
kind of element displayed.

Some navigational links are also added (to the index and to the
module index) to all documentation pages.

A refactoring of the DocumentationElement tree, the Processor and
templates was also done to ease the process.


jponge:
screen shot 2017-01-11 at 22 13 08

Absolutely gorgeous!

Issue #388, opened 2016-07-17
Tail call recursion support


jponge:
Tail call recursion is a classic compilation optimisation.

Since Golo encourages functional idioms when they make sense, it would be great to provide support for this optimisation.

yloiseau:
I'be been thinking about this one for a while.

jponge:
Sweet dreams. 
> 
> 
> 
>

PR #435, opened 2017-01-23
Experiment on TCE

yloiseau:
This is just a small experiment on tail call elimination (#388). This is preliminary work here for feed back.

Since we just identify tail call recursion with a `ReturnStatement` whose expression is a `FunctionCall` to the same function it pertains to, a lot of effectively TC constructions are not covered. For instance, none of the following functions will be optimized, despite being effectively tail recursive:
```golo
function foo = |acc, i| {
  if i <= 0 {
    return acc
  }
  let res = foo(acc + 1, i - 1)
  return res
}
```

The next step (the hard one) is to try to identify such cases and rewrite the IR to allow the optimization. I fear a lot of special cases that would lead to ugly code.
The other option would be to let the developer properly code the function, but this forbid interesting constructs (e.g. the `-> match` one).

Here is a snippet to “test” the feature:
```golo
module AsmTceStack

function blowup = |i| {
  if (i % 100) == 0 {
    println(i)
  }
  return blowup(i + 1)
}

function main = |args| {
  try {
    blowup(1)
  } catch (e) {
    println(e)
  }
}
```

when running it with
```bash
> GOLO_OPTS='-Xss256k -Dgolo.optimize.tce=false' golo golo --files asmtce-stack.golo
100
200
300
400
500
java.lang.StackOverflowError
```

using TCE, (`-Dgolo.optimize.tce=true` which is the default) will not overflow :tada:

## TODO:

- [x] deal with named arguments
- [x] deal with closed arguments (i.e. synthetic parameters)
- [x] deal with varargs: seams not doable...
- [x] deal with decorators: decorated functions are not optimized
- [x] documentation
- [x] tests

### DIFF
--- a/doc/decorators.adoc
+++ b/doc/decorators.adoc
@@ -1,4 +1,5 @@
 
+[[decorators]]
 == Decorators
 
 Golo features Python-like decorators.
--- a/doc/functions.adoc
+++ b/doc/functions.adoc
@@ -288,6 +288,183 @@ function i_will_crash = {
 }
 ----
 
+
+=== Recursive Tail Call Optimization
+
+Golo feature recursive https://en.wikipedia.org/wiki/Tail_call[tail call] optimization. If a function last action is to return the result of a recursive call, it is optimized to not stack a new call and is compiled in code equivalent to a loop. For instance, a function like:
+
+[source,golo]
+----
+function fact = |acc, v| -> match {
+  when v == 0 then acc
+  otherwise fact(acc * v, v - 1)
+}
+----
+
+is compiled into something roughly equivalent to :
+
+[source,golo]
+----
+function fact = |acc, v| {
+  var acc_ = acc
+  var v_ = v
+  while v_ != 0 {
+    acc_ = acc_ * v_
+    v_ = v_ - 1
+  }
+  return acc_
+}
+----
+
+This allows to create recursive functions that will not throw a `StackOverflowError` even in the presence of a large number or recursive call.
+The optimization can be disabled by setting the `golo.optimize.tce` system property to `false` (e.g. `export GOLO_OPTS='-Dgolo.optimize.tce=false`).
+
+A call is tail recursive when the function returns the result of calling itself directly. Indeed, since no evaluation nor flow analysis is done, many effectively tail recursive calls can't be identified as such, and thus are not optimized. It is recommended to rewrite the code to make the tail call more direct. For instance, the following two functions are not optimized:
+
+[source,golo]
+----
+function foo = |a, v| {
+  let r = ^foo
+  if v == 0 {
+    return a
+  }
+  return r(a + v, v - 1)
+}
+
+function bar = |a, v| {
+  var r = null
+  if v == 0 {
+    r = a
+  } else {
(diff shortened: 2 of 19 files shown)


yloiseau:
Ping @jponge :smile:

jponge:
I admire how you manage to both do crazy experiments _and_ welcome a lovely baby at home.
I didn't have that much energy 2 years back 😄 

I will of course review and test 👍 

I also added @Artpej

yloiseau:
I do crazy things instead of correcting exams... :-)

yloiseau:
I've an issue with varargs…

TCE works by storing the new arguments and `goto` the beginning of the function. For instance, in
```golo
function fact = |acc, v| -> match {
  when v == 0 then acc
  otherwise fact(acc * v, v - 1)
}
```
in the otherwise case, we store `acc * v` in register 0, `v - 1` in register 1, and return to the beginning, reloading 0 and 1 and so on.

```
 0: aload_1
 1: iconst_0
 2: invokestatic  #32       // Integer.valueOf
 5: invokedynamic #43,  0   // equals
10: checkcast     #45       // Boolean
13: invokevirtual #49       // booleanValue
16: ifeq          21
19: aload_0
20: areturn
21: aload_0
22: aload_1
23: invokedynamic #52,  0   // times
28: astore_2
29: aload_1
30: iconst_1
31: invokestatic  #32       // Integer.valueOf
34: invokedynamic #55,  0   // minus
39: astore_3
40: aload_2
41: astore_0
42: aload_3
43: astore_1
44: goto          0
```

Given
```golo
function f = |a, v...| -> match {
  when v: isEmpty() then f(0, 1, 2, a)
  when v: size() == 1 then f(v: get(0), a)
  otherwise f(a + v: head(), v: tail())
}
```

In the first case, we could store `0` in 0, and `array[1, 2, a]` in 1, then
`goto 0`. Fine. Similarly, in the second one, `v: get(0)` is store in 0, and
`array[a]` should be stored in 1. Ok. But in the last case, we should store `v: tail()` in 1 and not `array[v: tail()]`, since `v: tail()` *is* the array.

I see no way to do this at compile time, and so will not optimize TCE with varargs function. Does anyone see a way to resolve this?

jponge:
I can't believe we have spam on reviews...

Artpej:
What's about decorated fonctions? (😘 From FOSDEM)

yloiseau:
Only tests are missing, but I have no idea on how to test such a feature... any way to set the stack size in the code?

yloiseau:
Ok, seems we can pass a stack size to the `Thread` constructor. I'll try this

jponge:
Haven't started a proper review, but on small examples it works.

jponge:
I suggest having @k33g review.

He's magnificent at finding bugs 😇

yloiseau:
BTW, I'm experimenting with an "optimization visitor" that will rewrite IR to (hopefully) allow more cases for TCE by doing some (simple) flow analysis.

yloiseau:
ping @jponge @k33g

k33g:
@yloiseau I drop an 👁  this we

PR #440, opened 2017-02-13
Print a warning when unable to load a class

### LINKED ISSUE #439: Warning on unavailable import
Imports are resolved at run-time by trying to load the class and searching through its declared methods. When a class or method is not found, we continue with the next one, ignoring exceptions in order to try the next resolution method.

However, it can lead to difficult to catch bugs, for instance in the presence of a typo in the imported module/package name (see #405), since the raise exception is a `NoSuchMethodError`, which can indeed be raised by a lot of different mistakes (e.g. bad arity or argument types in the call, typo in the function name, etc.).

I propose to print a warning (disabled by default: `golo.warnings.unavailable-class=false`) when a module imports an unavailable module to help the diagnostic of such errors.

yloiseau:
FIX #439

When failing to load a class, either in a class reference or in a
function resolution (via imported modules or FQN call), a warning is
printed. This ease the debugging of typo or classpath errors.

The warning can be disabled by the `golo.warnings.unavailable-class`
system property.

### DIFF
--- a/doc/functions.adoc
+++ b/doc/functions.adoc
@@ -219,12 +219,12 @@ These modules are imported *after* the module explicitly imported in the module,
 
 [[warning-unavailable-class]]
 [WARNING]
-====================
-When an imported module can't be loaded, either because it is not in the classpath or because of a typo in its name, a warning is printed. This does not preclude to code to work correctly since the looked up function can be present in another imported module.
+====
+When an imported module can't be loaded, either because it is not in the classpath or because of a typo in its name, a warning is printed. This does not preclude the code to work correctly since the looked up function can be present in another imported module.
 When the function is called prefixed with the module name, and the module can't be loaded, the warning is also printed, but the code will fail with a `NoSuchMethodError`.
 
 The warning can be disabled by setting the `golo.warnings.unavailable-class` system property to `false`.
-====================
+====
 
 
 === Local functions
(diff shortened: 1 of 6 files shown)


doc/functions.adoc
```diff
@@ -217,6 +217,16 @@ Golo modules have a set of implicit imports:
 
 These modules are imported *after* the module explicitly imported in the module, so that elements defined in these modules (e.g. predefined functions or xref:_augmentations_resolution_order[augmentations]) can be redefined.
 
+[[warning-unavailable-class]]
+[WARNING]
+====================
+When an imported module can't be loaded, either because it is not in the classpath or because of a typo in its name, a warning is printed. This does not preclude to code to work correctly since the looked up function can be present in another imported module.
```
jponge: preclude _the_ code

doc/functions.adoc
```diff
@@ -217,6 +217,16 @@ Golo modules have a set of implicit imports:
 
 These modules are imported *after* the module explicitly imported in the module, so that elements defined in these modules (e.g. predefined functions or xref:_augmentations_resolution_order[augmentations]) can be redefined.
 
+[[warning-unavailable-class]]
+[WARNING]
+====================
```
jponge: Just 4 symbols will do.

PR #441, opened 2017-02-13
Localization of error messages and command line options

yloiseau:
I started to localize as much as I can. Not finished yet. 

I'm not sure if we should localize runtime exception messages (e.g. when resolving a function).

Maybe I'm doing it wrong, feel free to comment.

### DIFF
--- a/src/main/java/gololang/Messages.java
+++ b/src/main/java/gololang/Messages.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2012-2017 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
+ *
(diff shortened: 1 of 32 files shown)


jponge:
Do you fallback on missing keys?

It often happens that with localisation translations fall behind the primary language.

yloiseau:
If the defined locale does not have a corresponding translation resource, or if the key is not defined in the translation resource, the english message is used if it's what you mean

yloiseau:
But I just noted that the colors in the output make syntastic (vim syntax checker) to fail. I need a way to turn it of, ether by checking if output is a TTY (if feasible) or at least with a command line flag

yloiseau:
Are we OK with this one?

PR #443, opened 2017-02-28
Copyright year update

yloiseau:
The big yearly copyright update


yloiseau:
@jponge ok for the shell script or do you rather put it in `src/` ?

PR #458, opened 2017-04-10
Integrate a changelog file

yloiseau:
The format is based on [Keep a Changelog](http://keepachangelog.com/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,233 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased]
+### Added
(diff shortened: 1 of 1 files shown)


yloiseau:
@jponge The idea is that the file is a "curated list of notable changes". I don't think it can be created from git history. Maybe from github pull requests data. Indeed, I see it more like a summary of the merged pull requests, grouped by kind of changed, than a `git log`, updated by the contributor in each pull request.
It would decouple the history from the hosting service, and ease the publication of release news.

I just stumble upon [Keep a Changelog](http://keepachangelog.com/en/0.3.0/) while playing with [Leiningen](https://leiningen.org/) (looking for inspiration to improve `golo new`), and thought it could be a good idea :smile: If you think it's not needed, I'm fine.

jponge:
I completely agree that keeping a log is useful, it's just that the wiki
already does that fine and GitHub wikis are Git repositories so you don't
have big migration challenges if need ever be.

PR #464, opened 2017-04-30
Improve the new command

### LINKED ISSUE #437: `-` versus `.` when generating `pom.xml`
When Golo generate a maven project, if the name of the project is `hey-how-are-you-doing` the module name in `main.golo` becomes `hey.how.are.you.doing` but it does not change in the `pom.xml` in the 2 `<mainClass>hey-how-are-you-doing</mainClass>` sections, then we get an error when executing the application. cc @jponge

yloiseau:
- change the way filenames and module names are managed: If the given
  project name doesn't contain dots, it is converted to CamelCase for
  the module name and the file has a generic name; if it has dots, it is
  used as is for the module name, and the file is in a folder reflecting
  the package structure (FIX #437);
- refactor the project initialization: use a dedicated class, use Golo
  templates for the generated files;
- add a `--vcs` option to create an ignore file and init the repository:
  currently mercurial and git;
- define several profiles (app, lib) with the `--profile` option that
  generate different project structures;
- add several files and directories to the project structure (README,
  test, samples, ...)


yloiseau:
@k33g 
On upper case and lower case, it's what was suggested in the [#437](https://github.com/eclipse/golo-lang/issues/437#issuecomment-287761781) comment. The idea is that if you provide a dotted name (i.e. a fully qualified module name), you know what you do, so we don't touch it. If the name is not a valid module name (contains spaces, dash, ...), we convert it into a valid name using CamelCase.

On the `--group` option, I considered it, but thought it would be too mush. But if everyone is OK, I can add it. In the same way, I considered a `--force` option, since currently the command fails if the target directory exists,  but the behavior is unclear: should we overwrite the existing files, or just generate the missing ones, allowing to init an existing project (in this case, what if the sources are not in the right directory, etc.)

In Gradle, the option is `mainModule`, and is included in the `build.gradle` for the `app` profile.

jponge:
I'll review by Friday

src/main/java/org/eclipse/golo/cli/command/InitCommand.java
```diff
@@ -16,145 +16,138 @@
...
-import java.nio.charset.StandardCharsets;
-import java.util.LinkedList;
-import java.util.List;
+import java.nio.file.*;
+import java.util.*;
 
 import static gololang.Messages.message;
 import static gololang.Messages.info;
 
+// TODO: add project files:
```
jponge: Then turn it into issues than todos :-)

yloiseau: It was unclear for me if it's a wanted behavior. I'll remove the note and open an issue to allow discussion.

Artpej:
I just realised that Mercurial must be installed, then the build will fail because of some unit tests. May be it it the same for Git. 

What do you think about introduce an intergration test flag into the build to enable this kind of tests?

yloiseau:
Indeed, when the option to init the VCS is given, it's just a system call to the local program, so if `hg` and `git` are not installed on the test machine, the test will fail. It's OK on CI, but not necessary locally. 
You mean, in the test suite, check that `hg` and `git` can be found and abort the test otherwise, or a more elaborated technique?
Since the syscall wrapper class (`org.eclipse.golo.cli.command.VCS`) raise a `InitException` in case the call failed, maybe just catching it in the test would be enough…

jponge:
We should skip when `git` or `hg` are not available.

PR #467, opened 2017-05-03
Add a predefined function to concatenate strings

yloiseau:
For instance

```golo
let s = str("The answer", " is ", 2 * 21)
```
This is functionally equivalent to:
```golo
let s = ["The answer", " is ", 2 * 21]: join("")
```


yloiseau:
@jponge `Arrays::toString` doesn't have the same behavior:

```golo
println(java.util.Arrays::toString(array["The answer", " is ", 2 * 21])
#> [The answer,  is , 42]
println(str("The answer", " is ", 2 * 21))
#> The answer is 42
```

PR #468, opened 2017-05-03
Utilities for BigInteger and BigDecimal

yloiseau:
BigInteger and BigDecimal had a literal notation, but not a conversion
function (contrary to other numerical types).

The range object is also adapted to deal with ranges of BigInteger.


src/main/java/gololang/BigIntegerRange.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright (c) 2012-2016 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
```
jponge: '17

src/test/java/gololang/BigIntegerRangeTest.java
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright (c) 2012-2016 Institut National des Sciences Appliquées de Lyon (INSA-Lyon)
```
jponge: '17

yloiseau: oups... old branch :smile:

PR #470, opened 2017-05-10
Adapt the Javadoc to look like the golodoc one

yloiseau:
- add a link to source
- change style to match the golodoc style
- add highlight support in the generated javadoc


doc/stylesheet.css
```diff
@@ -0,0 +1,249 @@
+/* Javadoc style sheet */
+/*
+Overall document style
```
jponge: /*

yloiseau: can you be more precise? :smile:

jponge: you lack one on this line :-)

PR #472, opened 2017-05-11
Fix findbugs warnings

jponge:


src/main/java/gololang/Predefined.java
```diff
@@ -534,15 +534,18 @@ public static void textToFile(Object text, Object file, Object charset) throws T
       System.out.write(str.getBytes(encoding));
     } else {
       Path path = pathFrom(file);
-      if (path.getParent() != null) {
-        Files.createDirectories(path.getParent());
-      }
-      Files.write(
+      if (path != null) {
```
yloiseau: I don't get how `pathFrom` can return `null`...

src/main/java/gololang/Predefined.java
```diff
@@ -534,15 +534,18 @@ public static void textToFile(Object text, Object file, Object charset) throws T
...
+        if (parent != null) {
+          Files.createDirectories(parent);
+        }
+        Files.write(
           path,
           str.getBytes(encoding),
           StandardOpenOption.WRITE,
           StandardOpenOption.CREATE,
           StandardOpenOption.TRUNCATE_EXISTING);
+      }
```
yloiseau: no `else`? What if `path` is `null`? (event if I don't know in which case :smile:)

src/main/java/org/eclipse/golo/doc/FunctionDocumentation.java
```diff
@@ -170,25 +170,50 @@ public FunctionDocumentation parent(DocumentationElement p) {
...
         c = arity() < o.arity() ? -1 : 1;
-        if (c == 0) {
-          c = varargs && !o.varargs() ? 1
-              : varargs && o.varargs() ? 0
-              : -1;
-          if (c == 0) {
-            c = parent.compareTo(o.parent);
-          }
-        }
       }
```
yloiseau: Not sure here... Indeed it's dead code, but may be due to a bug. We are comparing functions. If arities are the same, we indeed should take varargs into account, no? I think the error is in the ternary operator comparing arity.

jponge: It was dead code since `could` only be `1` or `1` if I recall right.

src/main/java/gololang/Predefined.java
jponge: Good catch

src/main/java/org/eclipse/golo/doc/FunctionDocumentation.java
yloiseau: My point is that is should not be dead code :smile:
I rather write `c = Integer.compare(arity(), o.artity())` and thus the `c == 0` case is no more dead code. But I may be wrong. However, it's only used to sort the functions in the documentations, so no big deal. (nb: it seems that I'm the author of this infamy :sweat:)

jponge:
Well we can fix it back you know :-)

BTW Findbugs is quite picky on null-checks, it requires a local variable. I
can understand why from a theoretical POV, but still... :-D

src/main/java/org/eclipse/golo/doc/FunctionDocumentation.java
yloiseau: Doing some archeology, I introduced this while adding the doc index. The `equals` and `hashCode` *was there*! and the dead code was not. I see why I introduced it, failing to update the previous statement… but I have no clue why I removed `equals` and `hashCode` :sweat:

yloiseau:
just to be sure, is it because in a concurrent context, the attribute (or meth. result) can be made null between the test and the use?

jponge:
Yes, it's due to possible side-effects so the checker has no clue about the
outcome of 2 successive calls to an object method (in general),

PR #480, opened 2017-06-19
Fix arrayTypeOf(byte.class) yields a ClassNotFoundException

### LINKED ISSUE #479: arrayTypeOf(byte.class) yields a ClassNotFoundException
Everything is in the title. `arrayTypeOf(byte.class)` yields a `ClassNotFoundException` when one would expect it to return the Java equivalent of `byte[].class`

```
Exception in thread "main" java.lang.ClassNotFoundException: [Lbyte;
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at gololang.Predefined.arrayTypeOf(Predefined.java:613)
	at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:627)
	at org.eclipse.golo.runtime.FunctionCallSupport.fallback(FunctionCallSupport.java:195)
	at MyModule.main(main.golo:7)
	at org.eclipse.golo.cli.command.spi.CliCommand.callRun(CliCommand.java:53)
	at org.eclipse.golo.cli.command.GoloGoloCommand.execute(GoloGoloCommand.java:59)
	at org.eclipse.golo.cli.Main.main(Main.java:69)
```

Artpej:
Fixes #479

### DIFF
--- a/src/test/java/gololang/PredefinedTest.java
+++ b/src/test/java/gololang/PredefinedTest.java
@@ -6,7 +6,6 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
-
 package gololang;
 
 import org.testng.annotations.Test;
@@ -31,6 +30,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.*;
 
 import org.eclipse.golo.runtime.AmbiguousFunctionReferenceException;
+import org.testng.Assert;
 
 public class PredefinedTest {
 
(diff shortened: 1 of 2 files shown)


src/test/java/gololang/PredefinedTest.java
```diff
@@ -230,6 +230,13 @@ public void test_isArray() {
   @Test
   public void test_arrayOfType() throws ClassNotFoundException {
     assertThat((Class) Predefined.arrayTypeOf(Object.class), sameInstance((Class) Object[].class));
+    assertThat((Class) Predefined.arrayTypeOf(byte.class), sameInstance((Class) byte[].class));
+    try {
+      Predefined.arrayTypeOf(new Object());
+      Assert.fail();
+    } catch (Throwable t) {
+      assertThat((Class) t.getClass(), sameInstance((Class) AssertionError.class));
```
yloiseau: `Assert.fail()` will also raise an `AssertionError` 😄

Artpej: oups :thinking: , I don't know how i will catch this one...

yloiseau: why not just `@Test(expectedExceptions = AssertionError.class)`?

Artpej: I don't want create a test case juste for one assertion (seems to me not consistent with the test suite here), and use of `expect exception` for the last assertion can be confusing for the rest of the test case.

yloiseau: fine

PR #485, opened 2017-06-27
Adds local declaration to expressions

yloiseau:
This allows to create local variables in an expression to improve
reading and maintaining, while keeping expressions style.

For instance:

```golo
    function f = |lst| -> match {
      when s > 100 then "big"
      when s > 10 then "medium"
      when s == 0 then "empty"
      otherwise "small"
    } where {
      s = lst: size()
    }
```

or 

```golo
function sum = |lst| -> sum(0, lst) where {
  sum = |acc, lst| -> match {
    when lst: isEmpty() then acc
    otherwise sum(acc + head, tail) where {
      head, tail... = lst
    }
  }
}

```


jponge:
This is beautiful

yloiseau:
@artpej proposed to use `with` instead of `where` to not add a new keyword. What do you think?

jponge:
`with` is good to me as well

PR #489, opened 2017-06-28
FIX #471: property method priority

### LINKED ISSUE #471: Priority in fluent properties
We introduced fluent properties, that allows to call `setValue(v)` as just `value(v)` and `getValue()` or `isValue()` as `value()`. 
However, if a `value()` method exists, the property is called instead.

For instance, given:
```java
public class A {

  public String getValue() {
    return "getter";
  }

  public Integer value() {
    return 42;
  }
}
```
and
```golo
module Test

function main = |args| {
  println(A(): value())
}
```

`javac A.java && golo golo --files test.golo` will print `getter`, while `42` is expected

yloiseau:
As described in #471, the priority magic is used only if no regular
method exists.


jponge:
Merge?

PR #490, opened 2017-07-05
Fix Gitter badge URL by encoding the space

samueldg:
The current Gitter badge URL contains a space, which makes the
badge incorrectly display as a link (at least in GitHub's rendering).

Changing the space for '%20' fixes this.

Change preview:

## Before

gitter-badge-before

## After

gitter-badge-after


jponge:
Thanks!

PR #495, opened 2017-08-21
Build gardening

jponge:
* Removal of the `findbugs` plugin
* Removal of the local file publishing


yloiseau:
Why do you remove findbug?

jponge:
My opinion is that it hasn't proved very useful, and it's mostly forced us
to add exceptions or rework code to introduce extra variables because it
can't see though some natural Java code constructs.

yloiseau:
Ok

PR #497, opened 2017-08-28
FIX 496 : Add support for CodeSource URL for GoloClassLoader

### LINKED ISSUE #496: How to get the main golo file's location?
How can I get where the golo file containing the "main" function is? In Java, you can find where the JAR file is using one of the solutions posted here : https://stackoverflow.com/questions/15359702/get-location-of-jar-file but none of them work in Golo using `golo golo ...`

utybo:
This PR aims to fix #496 where it is impossible to get a .golo file's location at runtime. This uses the same mechanism as regular .jar and .classes in Java, which is registering a ProtectionDomain and CodeSource.

This PR adds a `load` method to `GoloClassLoader` that takes one argument more than the already present load method, which is a URL. This URL represents the original location of the .golo file.

Both the CLI "golo golo" and "golosh" were changed to use the new method. The old method was kept to ensure backwards-compatibility and still provide a way to load golo files without providing its source. This could also be useful in the future if you guys want to use Certificates or other security features of ProtectionDomain with GoloClassLoader.

### DIFF
--- a/src/main/java/org/eclipse/golo/cli/command/GoloGoloCommand.java
+++ b/src/main/java/org/eclipse/golo/cli/command/GoloGoloCommand.java
@@ -78,7 +78,7 @@ public class GoloGoloCommand implements CliCommand {
       }
     } else if (file.getName().endsWith(".golo")) {
       try (FileInputStream in = new FileInputStream(file)) {
-        Class<?> loadedClass = loader.load(file.getName(), in);
+        Class<?> loadedClass = loader.load(file.getName(), in, file.toURI().toURL());
         if (module == null || loadedClass.getCanonicalName().equals(module)) {
           return loadedClass;
         }
--- a/src/main/java/org/eclipse/golo/cli/command/ShebangCommand.java
+++ b/src/main/java/org/eclipse/golo/cli/command/ShebangCommand.java
@@ -78,7 +78,7 @@ public class ShebangCommand implements CliCommand {
     try (InputStream is = Files.newInputStream(path)) {
       Path filename = path.getFileName();
       if (filename != null) {
-        return loader.load(filename.toString(), is);
+        return loader.load(filename.toString(), is, path.toUri().toURL());
       } else {
         throw new RuntimeException(message("not_regular_file", path));
       }
--- a/src/main/java/org/eclipse/golo/compiler/GoloClassLoader.java
+++ b/src/main/java/org/eclipse/golo/compiler/GoloClassLoader.java
@@ -10,6 +10,10 @@
 package org.eclipse.golo.compiler;
 
 import java.io.InputStream;
+import java.net.URL;
+import java.security.CodeSource;
+import java.security.ProtectionDomain;
+import java.security.cert.Certificate;
 import java.util.List;
 
 /**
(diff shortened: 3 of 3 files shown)


yloiseau:
I think we should take the opportunity of this PR to refactor code loading in the commands. Indeed, almost every commands scan some directory for golo files and load/compile them, each in its own way.
We should refactor all this to have a single source scanning method, that yield `Path` objects, or even `CodeSource` object. 

The `GoloClassLoader::load` can be refactored to take only a single `CodeSource` argument, and derive the `goloSourceFilename` and `sourceCodeInputStream` from its `URL`.

src/main/java/org/eclipse/golo/compiler/GoloClassLoader.java
```diff
@@ -58,4 +62,25 @@ public GoloClassLoader() {
...
+   * This method declares a URL for the CodeSource of this class, which
+   * is useful for retrieving the location at runtime.
+   *
+   * @param goloSourceFilename    the source file name.
+   * @param sourceCodeInputStream the source input stream.
+   * @param sourceCodeLocation    the source file location.
+   * @return the class matching the Golo module defined in the source.
+   * @throws GoloCompilationException if either of the compilation phase failed.
+   */
+  public synchronized Class<?> load(String goloSourceFilename, InputStream sourceCodeInputStream, URL sourceCodeLocation) throws GoloCompilationException {
```
yloiseau: This method is copy/past from the previous one. You should refactor the previous one to call this one with a `null` location.
Or even better, as stated in the main comments, refactor the method to only take a `CodeSource` and derive the name and input stream from it, and remove the old one.

src/main/java/org/eclipse/golo/cli/command/GoloGoloCommand.java
```diff
@@ -78,7 +78,7 @@ public void execute() throws Throwable {
       }
     } else if (file.getName().endsWith(".golo")) {
       try (FileInputStream in = new FileInputStream(file)) {
-        Class<?> loadedClass = loader.load(file.getName(), in);
+        Class<?> loadedClass = loader.load(file.getName(), in, file.toURI().toURL());
```
yloiseau: Build the `CodeSource` instance here (see the proposed `load` refactoring). This will allows to construct a `Certificate[]` from command line options if we ever need to.
BTW, the file scanning facility should be extracted.

src/main/java/org/eclipse/golo/cli/command/ShebangCommand.java
```diff
@@ -78,7 +78,7 @@ private void loadOtherGoloFiles(GoloClassLoader loader, Path basedir, Path scrip
     try (InputStream is = Files.newInputStream(path)) {
       Path filename = path.getFileName();
       if (filename != null) {
-        return loader.load(filename.toString(), is);
+        return loader.load(filename.toString(), is, path.toUri().toURL());
```
yloiseau: idem

yloiseau (changes requested):

src/main/java/org/eclipse/golo/cli/command/GoloGoloCommand.java
utybo: @yloiseau what do you mean by "extracted"?

utybo:
Alright, I've seen your comments but am not sure of when I'll be able to work on this, though it seems to be pretty easy, so if anyone wants to do this instead of me feel free to do so.

src/main/java/org/eclipse/golo/cli/command/GoloGoloCommand.java
yloiseau: There should be a mode generic method that scan (recursively) a directory and return golo files, maybe in the form of a `CodeSource` object, or `Path` with a method to convert a `Path` into a `CodeSource`

yloiseau:
Since it's not a critical bug/feature, you can take some times to implement it :wink:

src/main/java/org/eclipse/golo/cli/command/GoloGoloCommand.java
yloiseau: see the "extract method" refactoring...

PR #516, opened 2018-03-05
Refactor ast/ir link

yloiseau:
Completely remove the link between the AST and the IR objects. This decouple the IR and the AST.
Also introduce the simulated self-type pattern to ease the definition of builder-like methods that returns the object itself on IR elements.

*NOTE*: this will probably break some IDE integration that works on / expect AST elements. However, as far as I can tell, these IDE plugins are already out of date and no more maintained (e.g. eclipse and netbeans ones still use the 'fr.insalyon.citi.golo' namespace). As a side note, maybe we should remove them from the documentation since they do not work anymore with the current golo version.

### DIFF
--- a/src/main/java/gololang/AdapterFabric.java
+++ b/src/main/java/gololang/AdapterFabric.java
@@ -77,8 +77,8 @@ public final class AdapterFabric {
       Object[] cargs = new Object[args.length + 1];
       cargs[0] = adapterDefinition;
       System.arraycopy(args, 0, cargs, 1, args.length);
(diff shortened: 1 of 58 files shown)


jponge:
I concur that references to the old IDE plugins should be removed.
The link between IR and AST had been introduced early on for the needs of the Netbeans plugin.

The future of language / editor / IDE integration is the _language server protocol_, clearly. Given resources it is clearly something we should develop, probably based on the server kit developed by the Eclipse IDE folks.

PR #517, opened 2018-03-07
FIX #513: issue with implicit upcast of primitive

### LINKED ISSUE #513: this simple program does not work
```golo
module teste.WebServer

import java.lang
import java.net.InetSocketAddress
import com.sun.net.httpserver.HttpServer

function main = |args| {

  let server = HttpServer.create(InetSocketAddress("localhost", 8081), 0)
	let handler1 = |exchange| {
    let response = "bem vindo"
    exchange: getResponseHeaders(): set("Content-Type", "text/plain")
		println(response: length())
    exchange: sendResponseHeaders(200, response: length())
    exchange: getResponseBody(): write(response: getBytes())
    exchange: close()
  }
  server: createContext("/", handler1)
	let handler2 = |exchange| {
	 let response = "tchau"
    exchange: getResponseHeaders(): set("Content-Type", "text/plain")
    exchange: sendResponseHeaders(200, response: length())
    exchange: getResponseBody(): write(response: getBytes())
    exchange: close()
    server: stop(5)
  }
  server: createContext("/shut", handler2)
  
  server: start()
  println("> http://localhost:8081/")
}
```

yloiseau:
When java method expect e.g. a float and it is called with a int, the
method is not found (since the big method resolution refactoring).

- adds a substituability comparison for method params
- use Extractors in function call resolution

### DIFF
--- a/src/main/java/org/eclipse/golo/runtime/Extractors.java
+++ b/src/main/java/org/eclipse/golo/runtime/Extractors.java
@@ -16,6 +16,7 @@ import java.lang.reflect.Modifier;
 import java.util.function.Predicate;
 
 import static org.eclipse.golo.runtime.TypeMatching.argumentsNumberMatches;
+import static org.eclipse.golo.runtime.TypeMatching.compareTypes;
 import static org.eclipse.golo.runtime.DecoratorsHelper.isMethodDecorated;
 
 public final class Extractors {
@@ -45,7 +46,7 @@ public final class Extractors {
         if (m2.isVarArgs() && !m1.isVarArgs()) {
           return -1;
         }
-        return 0;
+        return compareTypes(m1.getParameterTypes(), m2.getParameterTypes());
       });
   }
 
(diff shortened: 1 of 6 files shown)


jponge:
Nice, the following now works!

```golo
module Yo

function main = |args| {
  println("Sleep...")
  Thread.sleep(3000)
  println("Done")
}
```

jponge:
Works on methods, too:

```golo
module Yo

import java.util.Random

function main = |args| {
  let rand = Random()
  let stream = rand: longs(100)
  println(stream: filter(|n| -> n % 2 == 0): count())
}
```

jponge:
@yloiseau It's the method handle chain doing the upcast, right?

yloiseau:
@jponge the issue was that the method was not found, so I changed the lookup. The cast itself is done as usual by adapting the method handle.

jponge:
Ah I missed the adaptation, can you point me to it?

yloiseau:
By adaptation I just mean `handle.asType`...

jponge:
Ok so it's in the MH chain :-)

Artpej:
:+1:

PR #525, opened 2018-04-10
Fix bug with Windows relative URIs in doc

yloiseau:
Relative URIs in the doc under windows had backslash as separators.


jponge:
Thank you!

PR #538, opened 2019-09-24
FIX: conflict between golo module and java package

yloiseau:
The `gololang.ir` module containing fluent factory methods for IR nodes
and augmentations on said nodes is renamed `gololang.ir.DSL`.

Indeed, the IR nodes being in the `gololang.ir` java package, a
golo code that `import gololang.ir` had access to the IR nodes and the
fluent API, which was quite elegant.
Since the golo code is compiled after the java code, no conflict was
raised as long as we stayed inside the compiler itself for java code, or
in golo code.

However, trying to use classes from `gololang.ir` package inside 3rd
party java code raise an error since the `gololang.ir` class existed due
to the golo module.

The renaming fixes this aspect, requiring to explicitly import
`gololang.ir.DSL` in order to use the fluent API in golo code.

### DIFF
--- a/src/main/golo/gololang/ir/DSL.golo
+++ b/src/main/golo/gololang/ir/DSL.golo
@@ -16,7 +16,7 @@ to improve fluent IR building.
 See the [`gololang.ir`](../../javadoc/gololang/ir/package-summary.html) java
 package javadoc for documentation on the IR elements themselves.
 ----
-module gololang.ir
+module gololang.ir.DSL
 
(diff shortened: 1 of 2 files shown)


jponge:
BTW you may need to fix the Git commit email for the Eclipse CLA webhook.

yloiseau:
Should we consider this a breaking change and increment the version number of the next release accordingly ?

jponge:
Absolutely

Issue #542, opened 2020-08-25
Upgrade to Gradle 6


jponge:
Golo is using an old version of Gradle. The upgrade to Gradle 6 is not straightforward, but we may benefit from doing so. We may also benefit from moving to the Kotlin DSL over the error-prone Groovy one.

PR #544, opened 2020-08-25
FIX #543: Upgrade some dependencies

### LINKED ISSUE #543: Upgrade some dependencies
Seek and upgrade some of the project dependencies where it makes sense.

yloiseau:
Upgrade some dependencies (#543)

- no change in JSONAssert since 2017 :scream:
- txtmark and JSONSimple has not change in a while, but other project share the name, not sure if its forks, we need to check
- test-logger >2.0 causes errors, upgrade to 1.7.1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,13 +86,13 @@ configurations.all {
 }
 
 dependencies {
-  compile 'org.ow2.asm:asm:7.0'
+  compile 'org.ow2.asm:asm:8.0'
   compile 'com.beust:jcommander:1.60'
   compile 'com.github.rjeschke:txtmark:0.13'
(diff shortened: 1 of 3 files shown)


jponge:
test-logger is probably better with a recent Gradle version (#542)

Perhaps you should just rewrite a modern Gradle build from scratch in Kotlin DSL, the migration could be easier. We weren't doing any wizardry in the current build, so it should be quite ok to reproduce most of the build steps.

yloiseau:
I agree. I just made the obvious changes while changing code dependencies

yloiseau:
Closing since changes are integrated in #545

PR #551, opened 2020-09-26
Adds Java annotation to golo elements

yloiseau:
Allows to annotate golo functions or types with Java annotations:
- generates bytecode for the annotations based on metadata attached to IR elements
- provides utility functions to ease the creation of macro adding such metadata
- provides a "meta-macro" generating such macros using reflexion on the annotation class
- define a `deprecated` macro to mark elements as deprecated (annotating with `java.lang.Deprecated` is not enough, the bytecode flag needs to be set)

### DIFF
--- a/src/main/golo/decorators.golo
+++ b/src/main/golo/decorators.golo
@@ -150,7 +150,7 @@ function lengthIs = |l| -> |v| {
 
 # .. Checkers .. #
 
-#XXX: checkers defined here need to be functions returning the actuall
+#XXX: checkers defined here need to be functions returning the actual
 # checkers, otherwise they are not accessible on import. Not very elegant.
 # Must be a better solution.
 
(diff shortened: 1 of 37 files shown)


src/main/golo/decorators.golo
```diff
@@ -150,7 +150,7 @@ function lengthIs = |l| -> |v| {
 
 # .. Checkers .. #
 
-#XXX: checkers defined here need to be functions returning the actuall
+#XXX: checkers defined here need to be functions returning the actual
```
jponge, line 153: Perhaps replace `XXX` :-)

jponge (approved):
Oh this is nice work :-)

src/main/golo/decorators.golo
yloiseau: This is a classical comment flag (see e.g. in highlight.js), but OK

Issue #548, opened 2020-08-28
Write tutorials


jponge:
- quick start
- use existing library
- ...

kelcecil:
Apologies for the accidental issue edit. I was browsing the issues on mobile and accidentally checked an item on the checklist when scrolling down the page.

PR #555, opened 2020-10-04
📝 how to start quickly with Golo

k33g:
This new tutorial will allow **beginners** to get comfortable quickly

Related issue https://github.com/eclipse/golo-lang/issues/548


k33g:
👋 @yloiseau @jponge the second one

jponge:
Nice, and... there are emojis!

tutorials/first-steps.md
```diff
@@ -0,0 +1,397 @@
...
+```golo
+module hello.world # 1️⃣
+
+function main = |args| { # 2️⃣
+  println("Hello world!")
+}
+```
+
+> - 1️⃣ always start with a module name
+> - 2️⃣ a Golo program has alaways a `main` function, and you use the pipe notation to pass the arguments
```
yloiseau, line 31: to declare the parameters... ;)

tutorials/first-steps.md
```diff
@@ -0,0 +1,397 @@
...
+  println(hello())
+  println(hello_who("Bob"))
+  println(hello_people("Bob", "Bill"))
+}
+
+```
+
+> - 0️⃣ you can use multi-lines comments at the header of the functions
+> - 1️⃣ this is a parameter-less function
+> - 2️⃣ you need to use the pipe notation to pass parameter(s)
```
yloiseau, line 146: to declare the parameters...

tutorials/first-steps.md
```diff
@@ -0,0 +1,397 @@
...
+  ) # 4️⃣ it will print 42
+
+}
+```
+
+> - 1️⃣ it will print `Hello`
+> - 2️⃣ it will print `Hello Bob`
+> - 3️⃣ Golo supports a compact form of closures for the cases where their body consists of a single expression
+> - 4️⃣ it will print `42`
+
```
yloiseau, line 194: `--pedantic` these are not closures, since no free variable is closed from the context of the lambda definition.
Besides the semantic discussion, maybe a example of a true closure could be useful, e.g.
```golo
  let adder = |first| -> |second| -> first + second

  let add_two = adder(2)
  let add_four = adder(4)

  println(add_two(40))
  println(add_four(38))
```
or maybe clearer written
```golo
let adder = |first| {
  return |second| -> first + second
}
```

tutorials/first-steps.md
```diff
@@ -0,0 +1,397 @@
...
+        return "NOT IN THE LIST"
+      }
+    }
+  }
+
+  println (choice("two")) # 2️⃣
+  println (choice("three")) # 3️⃣
+
+  # The values to be returned are specified after a then keyword that follows a boolean expression to be evaluated.
+
```
yloiseau, line 241: you're mixing `case` and `match`

tutorials/first-steps.md
```diff
@@ -0,0 +1,397 @@
...
+  let value = 5
+
+  if value <= 5 { # 1️⃣
+    println("value is <= 5")
+  } else {
+    println("value is > 5")
+  }
+
+  # the case construction can only be used inside a closure or a function
```
yloiseau, line 222: you can use the case as a statement: 
```golo
let value = "one"
case {
    when value == "one" {
        println("ONE")
    }
    when value == "two" {
        println("TWO")
    }
    otherwise {
        println("NOT IN THE LIST")
    }
}
```

it's just a `if ... else if ... else if ... else` (compiled as such)

yloiseau (changes requested):

tutorials/first-steps.md
k33g: it comes from the Golo documentation

tutorials/first-steps.md
k33g: 🤔 I'm not sure to understand the remark

k33g: 😮 oh I see ..

jponge:
👍

tutorials/first-steps.md
yloiseau: Yes, the "closure" term is a bit overloaded :wink:

yloiseau (approved):

PR #562, opened 2020-10-19
Use a fancy test logger

jponge:


yloiseau:
Color scheme does not works well on my solarized dark terminal...

screenshot-1603139191

jponge:
What would work better? See https://github.com/radarsh/gradle-test-logger-plugin/blob/develop/docs/SCREENSHOTS.md

yloiseau:
Standard theme is OK, but less fancy. 

screenshot-1603184062

On the other hand, Mocka is OK on light background. It may be a bug on their side, or event in my terminal theme configuration.
screenshot-1603183888

jponge:
Just let me know, we can change the theme to standard if it sounds safer to you. I'm not on Solarized but Dracula, works great.

yloiseau:
Ok, go on, I'll adapt :smile:

PR #566, opened 2020-10-25
Reworked Dockerfile

jponge:


k33g (approved):
👍 🧹🐳

Dockerfile
```diff
@@ -1,14 +1,9 @@
...
-RUN cd /src &&\
-  ./gradlew -v &&\
-  ./gradlew installDist &&\
-  mkdir -p /opt/golo &&\
-  cp -R /src/build/install/golo/* /opt/golo &&\
-  ln -s /opt/golo/bin/golo /usr/bin/golo &&\
-  rm -rf /src /root/.gradle
+FROM azul/zulu-openjdk-alpine:8
+COPY --from=builder /src/build/install/golo /opt/golo
+RUN ln -s /opt/golo/bin/golo /usr/bin/golo
```
yloiseau, line 8: Why don't you add to PATHs instead of symlink ?
- `/opt/golo/man` into `MANPATH`
- `/opt/golo/bin` and maybe `/opt/golo/share/shell-completion/` into `PATH`
- maybe `opt/golo/lib/*.jar` to `CLASSPATH`

yloiseau (approved):

Dockerfile
jponge: Good idea

PR #567, opened 2020-10-30
Sort compiling files

yloiseau:
Macros can introduce compile time dependencies, since a macro providing module must be available when compiling a module using it.
This PR refactors the commands to sort the compiled files according to dependencies.
In addition, the module containing a `main` function is sorted last to allow execution.


src/main/java/org/eclipse/golo/cli/GolofilesManager.java
```diff
@@ -0,0 +1,199 @@
...
+import org.eclipse.golo.compiler.CodeGenerationResult;
+import org.eclipse.golo.cli.command.Metadata;
+import static gololang.Messages.*;
+
+/**
+ * Helper class to deal with Golo files.
+ *
+ * <p>Ease the finding, loading and saving of golo source files and compilation result.
+ */
+public final class GolofilesManager implements AutoCloseable, Consumer<CodeGenerationResult> {
```
jponge, line 39: Would be better as `GoloFilesManager`

jponge (approved):

PR #575, opened 2020-11-29
Python-like Contexts

yloiseau:
Contexts are an improved form of try-with-resource.

### DIFF
--- a/src/main/golo/gololang/Control.golo
+++ b/src/main/golo/gololang/Control.golo
@@ -1,0 +1,497 @@
+----
+This module contains utilities to control execution flow.
+
+## Contexts
+
+Contexts are very similar to Python's
+[`with` statement](https://docs.python.org/3/reference/compound_stmts.html#with)
+and can be seen as a more generic version of Java's
+[try with resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)
+
+A *context* is any object exposing two special methods.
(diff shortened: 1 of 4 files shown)


jponge (approved):
This is impressive work @yloiseau 👏  👏  👏

k33g:
@yloiseau 🙏 can you please sync your feature branch with `master`, so I will be able to build it with GitPod to play with contexts

yloiseau:
@k33g done!

k33g (approved):
build ✅

my test:

```golo
module my.context

import gololang.Control

function main = |args| {
  # enter: a closure executed when entering the context.
  # This closure takes no argument and return the value that will be bound to the context variable.
  let enter = -> {
    # do something
    println("👋 hey I'm the enter closure")
    return 42/0
  }

  # exit: a closure executed when exiting the context.
  # This closure has two parameters: the target (as returned by `enter`) and the exception raised by the wrapped block.
  let exit = |targetValueFromEnter, raisedException| {
    case {
      when raisedException is null {
        println("😃 all is ok 👏, the value is: " + targetValueFromEnter)
      }
      otherwise {
        println("😡 something wrong...")
        return raisedException
      }
    }
  }

  let myContext = context(enter, exit)

  &within(thing = myContext) {
    println("🤖 result is: " + thing)
    println("🤖 result is: " + thing())
  }

}

```
All good for me 👏  @yloiseau

